PR #3, opened 2022-09-14
Add `BertConfig` and `RobertaConfig` for hyperparameters

shadeMe:


curated_transformers/models/bert/config.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class BertConfig:
```
danieldk, line 5: I wonder if it makes sense to split this up into more than once class. E.g., we could have a `BertEmbeddingConfig` and `BertLayerConfig` and these would then be member of `BertConfig`. Then downstream classes could have different combinations of things. Like at some point we could have a model that uses `BertLayerConfig` and `SinusoidalEmbeddingsConfig`. And the model-specific encoder class could just pass through the sub-configuration.

danieldk: In this setup, there would also be fields in the `Config` class itself. E.g. an ALBERT model would use `BertLayerConfig` for the layers, but it's own configuration of the number of layers and the layer grouping.

shadeMe: Ah, very good point. That's definitely more flexible that this proposal; will implement.

shadeMe:
The split-implementation introduces some redundancy since some of the hyperparameters are shared between different model components, but it's more composable overall.

danieldk (approved):

PR #12, opened 2022-09-26
Average windows overlap in spans

danieldk:
Suppose we have:

```
A B C D E F G H I
<--->   stride
<-----> window
```

D is part of both the current and the next window. To ensure that there is no drift in representations, we average them in both representations.

### DIFF
--- a/curated_transformers/models/with_strided_spans.py
+++ b/curated_transformers/models/with_strided_spans.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple
+from typing import Callable, List, Tuple
 from functools import partial
 from thinc.model import Model
 from thinc.types import Ragged, Floats2d, Ints1d
@@ -11,6 +11,11 @@ def build_with_strided_spans_v1(stride=96, window=128):
 def with_strided_spans(
     layer, *, stride=96, window=128
 ) -> Model[List[Ragged], List[Ragged]]:
+    if not (window // 2 <= stride <= window):
+        raise ValueError(
+            f"Stride must be within [window / 2, window] ([{window // 2}, {window}]), was: {stride}"
+        )
+
     attrs = {
         "stride": stride,
         "window": window,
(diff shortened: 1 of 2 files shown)


curated_transformers/models/with_strided_spans.py
```diff
@@ -51,25 +51,65 @@ def with_strided_spans_forward(
...
+    Y_docs = _strided_arrays_to_ragged(
+        model, Y_layer, doc_lens, stride=stride, window=window
+    )
+
     def backprop(dY):
         dY_spans, dY_lengths = _ragged_to_strided_arrays(
             dY, stride=stride, window=window
         )
         dXlf = backprop_layer(dY_spans)
+        _average_overlaps(model, Y_layer, doc_lens, stride=stride, window=window)
```
danieldk, line 64: Still in draft. Because I am not quite sure yet if we want to do this in backprop. Plus the averaging should probably happen before backprop'ing into the wrapped layer?

danieldk: By the way, I just realized that we are backproping for the overlapping outputs twice? `window - stride` `dY`s are used for both the end of the current window and the start of the next window. I am not sure whether this is fine or not?

danieldk: We discussed this a bit and I greed it would be best to use the mathematically-correct (?) approach of backprop'ing twice with half the gradient (since on the forward we do `(u + v) / 2`). Empirically there doesn't seem to be much of a difference between different gradients, as long as we are averaging the overlaps.

curated_transformers/models/with_strided_spans.py
```diff
@@ -51,25 +56,92 @@ def with_strided_spans_forward(
...
+    window: int,
+    func: Callable,
+):
+    """Average representations of overlapping windows. This function
+    modifies the arrays in Xlf in-place to avoid"""
+
+    if stride > window:
+        raise ValueError(
+            f"Stride size ({stride}) should less than or equal to window size ({window})"
+        )
```
shadeMe, line 113: Could we move this check to the entry-point, next to `if window // 2 > stride` or merge them (since we don't expect the value to change after model construction)?

curated_transformers/models/with_strided_spans.py
```diff
@@ -51,25 +56,92 @@ def with_strided_spans_forward(
...
+def _apply_to_overlaps(
+    Xlf: List[Floats2d],
+    lens: List[Ints1d],
+    *,
+    stride: int,
+    window: int,
+    func: Callable,
+):
+    """Average representations of overlapping windows. This function
+    modifies the arrays in Xlf in-place to avoid"""
```
shadeMe, line 108: Missing a word after `avoid`.

curated_transformers/models/with_strided_spans.py
```diff
@@ -51,25 +56,92 @@ def with_strided_spans_forward(
...
+
+def _normalize_gradients(array1, array2):
+    array1 *= 0.5
+    array2 *= 0.5
+
+
 def _ragged_to_strided_arrays(
     Xlr: List[Ragged], *, stride: int, window: int
 ) -> Tuple[List[Floats2d], List[List[int]]]:
+
```
shadeMe, line 144:
```suggestion
```

curated_transformers/tests/models/test_with_strided_spans.py
```diff
@@ -16,6 +16,16 @@ def backprop(dY: Floats2d) -> Floats2d:
     return Model("relu_activation", forward)
 
 
+def _add_range() -> Model[Floats2d, Floats2d]:
+    """Add range [0, X.size]."""
```
shadeMe, line 20:
```suggestion
    """Add range [0, X.size)."""
```

curated_transformers/tests/models/test_with_strided_spans.py
```diff
@@ -50,3 +60,22 @@ def test_with_strided_spans():
     ops.xp.testing.assert_array_equal(dX[1].data, zeros)
     ops.xp.testing.assert_array_equal(dX[0].lengths, lengths1)
     ops.xp.testing.assert_array_equal(dX[1].lengths, lengths2)
+
+
+def test_with_strided_spans_averaging():
+    ops = NumpyOps()
+    stateful = with_array(_add_range())
+    model = with_strided_spans(stateful, stride=2, window=4)
```
shadeMe, line 68:
```suggestion
	with pytest.raises(ValueError):
		model = with_strided_spans(stateful, stride=2, window=6)
	with pytest.raises(ValueError):
		model = with_strided_spans(stateful, stride=4, window=3)

    model = with_strided_spans(stateful, stride=2, window=4)
```

curated_transformers/models/with_strided_spans.py
danieldk: Moved to the entry point.

curated_transformers/models/with_strided_spans.py
danieldk: Not sure what I wanted to say there, so I just put a period after _in-place_ now.

curated_transformers/models/with_strided_spans.py
danieldk: Fixed.

curated_transformers/tests/models/test_with_strided_spans.py
danieldk: Good catch!

curated_transformers/tests/models/test_with_strided_spans.py
danieldk: Added, but moved to a separate test.

curated_transformers/models/with_strided_spans.py
```diff
@@ -11,6 +11,16 @@ def build_with_strided_spans_v1(stride=96, window=128):
...
 ) -> Model[List[Ragged], List[Ragged]]:
+    if stride > window:
+        raise ValueError(
+            f"Stride size ({stride}) should less than or equal to window size ({window})"
+        )
+
+    if window // 2 > stride:
+        raise ValueError(
+            f"Stride ({stride}) should at least be half the size of the window ({window})"
+        )
```
shadeMe, line 22: Let's merge these two into a single bounds check.

danieldk: Fixed.

shadeMe (approved):

PR #14, opened 2022-09-29
Remove dependency on 🤗 transformers

danieldk:
Before we this change, we would always load the model weights using the `transformers` library, even when deserializing finetuned models. This change postpones model loading to the `init` callback of the piece and transformer encoder models. In this way, we only attempt to load models when a new pipeline is initialized.

To further remove the dependency on the `transformers` library, we make the actual loader configurable through the new `model_loaders` registry. This allows us to provide other loaders in the future. This is an example configuration of a transformer pipe using the model loader mechanism:

```
[components.transformer]
factory = "curated_transformer"

[components.transformer.model]
@architectures = "curated-transformers.XLMRTransformer.v1"
vocab_size = 250002

[components.transformer.model.encoder_loader]
@model_loaders = "curated-transformers.HFEncoderLoader.v1"
name = "xlm-roberta-base"

[components.transformer.model.piecer_loader]
@model_loaders = "curated-transformers.HFSentencepieceLoader.v1"
name = "xlm-roberta-base"

[components.transformer.model.with_spans]
@architectures = "curated-transformers.WithStridedSpans.v1"
stride = 96
window = 128
```

Since we cannot rely on the `transformers` library anymore to provide a model's hyperparameters, we use the default hyperparameters for a model instead. The configuration should specify deviations from the default configuration, typically the vocab size, hidden size, number of layers, and number of attention heads.

### DIFF
--- a/curated_transformers/models/hf_wrapper.py
+++ b/curated_transformers/models/hf_wrapper.py
@@ -1,29 +1,27 @@
-import re
-import torch
-
-from typing import Tuple, List, Dict
+from typing import Tuple, List, Dict, Callable
 from functools import partial
 
 from spacy.tokens import Doc
 from thinc.api import Model, torch2xp, get_current_ops
 from thinc.api import xp2torch, PyTorchWrapper_v2
+from thinc.model import empty_init
 from thinc.shims.pytorch_grad_scaler import PyTorchGradScaler
 from thinc.types import ArgsKwargs, Floats2d, Ints1d, Ragged
 
 from .bert import BertEncoder, BertConfig
 from .roberta import RobertaEncoder, RobertaConfig
 from .._compat import transformers, has_hf_transformers
+from ..util import registry
 from .hf_util import convert_hf_pretrained_model_parameters
 
 
 def build_hf_transformer_encoder_v1(
     encoder: Model[List[Doc], List[Ragged]],
+    *,
     mixed_precision: bool = False,
     grad_scaler_config: dict = {},
+    init: Callable = empty_init,
 ) -> Model[List[Ints1d], List[Floats2d]]:
-    if not has_hf_transformers:
-        raise ValueError("requires 🤗 transformers")
-
     if "enabled" not in grad_scaler_config:
         grad_scaler_config["enabled"] = mixed_precision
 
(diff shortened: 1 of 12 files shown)


curated_transformers/models/hf_wrapper.py
```diff
@@ -38,6 +36,10 @@ def build_hf_transformer_encoder_v1(
         mixed_precision=mixed_precision,
         grad_scaler=PyTorchGradScaler(**grad_scaler_config),
     )
+
+    model.init = init
+    model.attrs["encoder"] = encoder
```
shadeMe, line 41: Do we want to store this in `models.refs` instead (as it's technically a sub-layer and we use that convention in `transformer_model`)?

curated_transformers/models/hf_wrapper.py
```diff
@@ -1,29 +1,27 @@
...
 from thinc.api import Model, torch2xp, get_current_ops
 from thinc.api import xp2torch, PyTorchWrapper_v2
+from thinc.model import empty_init
 from thinc.shims.pytorch_grad_scaler import PyTorchGradScaler
 from thinc.types import ArgsKwargs, Floats2d, Ints1d, Ragged
 
 from .bert import BertEncoder, BertConfig
 from .roberta import RobertaEncoder, RobertaConfig
 from .._compat import transformers, has_hf_transformers
+from ..util import registry
```
shadeMe, line 14: Seems like the `util` module is not part of this changeset?

danieldk: Oops, added.

curated_transformers/models/hf_wrapper.py
danieldk: The PyTorch wrapper doesn't add the wrapped model to the `layers` member. The model is wrapped in a shim and added to `shims`. That probably makes sense, because it's not a Thinc model. But I think using `attrs` is probably not good either, because I think the decode is not reentrant after deserialization. Maybe we should just access the shim directly where needed?

danieldk: Fixed. It's slightly ugly, since we are looking at the internals of the shim, but I don't see any better solution.

curated_transformers/util.py
```diff
@@ -0,0 +1,5 @@
+from importlib.metadata import entry_points
```
shadeMe, line 1: What does this import do?

shadeMe, line 1: What does this import do? We're not explicitly using in the module.

shadeMe (approved):

PR #15, opened 2022-09-30
Add an example project for training an XLM-R model

danieldk:


project/project.yml
```diff
@@ -0,0 +1,78 @@
+title: "Biaffine Parser"
```
shadeMe, line 1:
```suggestion
title: "Transition-based Parser"
```

danieldk: Changed the title to something more descriptive 👍 .

shadeMe (approved):

PR #16, opened 2022-09-30
Add option to freeze `TransformerPipe`

shadeMe:


curated_transformers/pipe.py
```diff
@@ -28,8 +28,10 @@
     assigns=["doc._.trf_data"],
     default_config={"model": DEFAULT_TRANSFORMER_MODEL},
 )
-def make_transformer(nlp: Language, name: str, model: Model) -> "Transformer":
-    return Transformer(nlp.vocab, model, name)
+def make_transformer(
+    nlp: Language, name: str, model: Model, frozen: bool = False
```
danieldk, line 32: I wonder if we should

```suggestion
    nlp: Language, name: str, model: Model, *, frozen: bool = False
```

to make sure that if someone uses this factory method directly, they do not rely on positional arguments.

curated_transformers/pipe.py
```diff
@@ -280,28 +290,36 @@ def update(
...
         docs = [eg.predicted for eg in examples]
         set_dropout_rate(self.model, drop)
-        tokvecs, bp_tokvecs = self.model.begin_update(docs)
-        d_tokvecs = [
-            Ragged(self.model.ops.alloc_f(t2v.dataXd.shape), t2v.lengths)
-            for t2v in tokvecs
-        ]
         losses.setdefault(self.name, 0.0)
 
+        if not self.frozen:
```
danieldk, line 295: Since we have to handle both cases anyway, maybe switch them around, so that we can have

```
if self.frozen:
```

curated_transformers/pipe.py
```diff
@@ -280,28 +290,36 @@ def update(
...
+            tokvecs, bp_tokvecs = self.model.begin_update(docs)
+            d_tokvecs = [
+                Ragged(self.model.ops.alloc_f(t2v.dataXd.shape), t2v.lengths)
+                for t2v in tokvecs
+            ]
+        else:
+            # Ensures that the inner torch model is executed in a `no_grad` context.
+            tokvecs = self.model.predict(docs)
+            bp_tokvecs = None
```
danieldk, line 304: My worry with the current implementation is: what if someone runs update when `frozen=True`, but then toggles `frozen=False` before the listeners backprop. This will then fail, because `d_tokvecs` is not allocated, etc.

I think `accumulate_gradient` and `backprop` below should be stateless (at least wrt. `self.frozen`). Instead, the listeners should receive a function/lambda that doesn't do anything when `self.frozen==True`. Then if someone switches to `frozen=False`, nothing bad happens when the listeners backpropagate.

danieldk (approved):

PR #20, opened 2022-10-05
Add support for post-training dynamic quantization

danieldk:
This change adds the following:

- `TorchScriptShim`/`TorchScriptWrapper_v1` to wrap TorchScript.
- A spaCy CLI extension that provides the `quantize-transformer`subcommand.

Besides that there are some modifications to the models:

- Use a constant for the attention mask so that the encoder can be converted to TorchScript. For some reason, this can also not be a global constant, so we have to inline the constant.
- Annotate `TransformerEncoderOutput` with `torch.jit.script`, so that it works with TorchScript.
- Extend the model factories with an `torchscript` option to force construction of TorchScript wrappers.

### DIFF
--- a/curated_transformers/cli/quantize.py
+++ b/curated_transformers/cli/quantize.py
@@ -1,0 +1,97 @@
+from io import BytesIO
+from pathlib import Path
+import spacy
+from spacy import Language
+from spacy.cli import app
+from thinc.api import Model, PyTorchShim
+import torch
+from torch.nn import Embedding, Linear, Module
+from torch.quantization import qconfig
+from typer import Argument as Arg, Option
+
+from ..models.torchscript_wrapper import to_torchscript_wrapper
+from ..pipe import Transformer
+
+
+@app.command("quantize-transformer")
+def quantize_cli(
+    model_path: Path = Arg(..., help="Model to quantize", exists=True, allow_dash=True),
+    output_path: Path = Arg(..., help="Output directory to store quantized model in"),
+    skip_embeds: bool = Option(
+        False, "--skip-embeds", help="Do not quantize embeddings"
+    ),
+    skip_linear: bool = Option(
+        False, "--skip-linear", help="Do not quantize linear layers"
+    ),
+):
+    """
+    Quantize a curated-transformers model.
+    """
+    nlp = spacy.load(model_path)
+    nlp_quantize_dynamic(nlp, skip_embeds=skip_embeds, skip_linear=skip_linear)
+    nlp.to_disk(output_path)
+
+
+def size_of_model(model):
+    filelike = BytesIO()
+    torch.save(model.state_dict(), filelike)
+    return filelike.tell()
+
+
+def nlp_quantize_dynamic(
+    nlp: Language, *, skip_embeds: bool = False, skip_linear: bool = False
+):
+    for pipe_name, pipe in nlp.components:
+        # We only quantize curated transformers, other models may not
+        # be able to deal with quantization and/or TorchScript.
+        if not isinstance(pipe, Transformer):
+            continue
+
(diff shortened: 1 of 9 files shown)


curated_transformers/models/torchscript_wrapper.py
```diff
@@ -0,0 +1,148 @@
+from typing import Any, Callable, Optional, Union
```
shadeMe, line 1: Would this module be a good candidate to add/move to `thinc`?

curated_transformers/cli/quantize.py
```diff
@@ -0,0 +1,97 @@
+from io import BytesIO
+from pathlib import Path
+import spacy
+from spacy import Language
+from spacy.cli import app
+from thinc.api import Model, PyTorchShim
+import torch
+from torch.nn import Embedding, Module
```
shadeMe, line 8:
```suggestion
from torch.nn import Module
```
Unused.

curated_transformers/cli/quantize.py
```diff
@@ -0,0 +1,97 @@
...
+):
+    for pipe_name, pipe in nlp.components:
+        # We only quantize curated transformers, other models may not
+        # be able to deal with quantization and/or TorchScript.
+        if not isinstance(pipe, Transformer):
+            continue
+
+        model = pipe.model.get_ref("transformer")
+
+        if model.name != "pytorch":  # Probably already quantized.
```
shadeMe, line 52: We could be explicit here and compare it to `pytorch_script` instead.

curated_transformers/models/torchscript_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+        self.cfg = msg["config"]
+        filelike = BytesIO(msg["model"])
+        filelike.seek(0)
+        self._model = torch.jit.load(filelike, map_location=device)
+        self._model.to(device)
+        self._grad_scaler.to_(device)
+        return self
+
+
+def TorchScriptWrapper_v1(
```
shadeMe, line 68: Would it make sense to register this entry-point in `setup.cfg`?

curated_transformers/models/torchscript_wrapper.py
danieldk: Yeah, this should eventually go into Thinc, once we are happy with the class.

shadeMe: I'll open an issue to keep track of it.

project/project.yml
```diff
@@ -61,18 +62,31 @@ commands:
...
       - "python -m spacy evaluate --gpu-id ${vars.gpu} training/${vars.treebank}/model-best corpus/${vars.treebank}/test.spacy"
+      - "python -m spacy evaluate training/${vars.treebank}/model-best-quantized corpus/${vars.treebank}/test.spacy"
     deps:
       - "training/${vars.treebank}/model-best"
+      - "training/${vars.treebank}/model-best-quantized"
       - "corpus/${vars.treebank}/test.spacy"
 
   - name: "evaluate-dev"
-    help: "Evaluate the parser model on the test corpus."
+    help: "Evaluate the pipelinel on the dev corpus."
```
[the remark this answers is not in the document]
danieldk: Fixed.

curated_transformers/cli/quantize.py
danieldk: Fixed, but slightly differently. I also import `Linear` now and replaced the fully qualified `torch.nn.{Embedding,Linear}` later in the file.

curated_transformers/cli/quantize.py
danieldk: 👍 much better

shadeMe (approved):

PR #22, opened 2022-10-07
quantize: add `max-mse-loss` option

danieldk:
This option excludes parameters from quantization for which the MSE would be larger than the given value.


danieldk:
> We should also update the `spacy project` to include the correct threshold for XLM-R (and a couple of comments clarifying the need).

Added to the project. I called the variable `max_mse_qloss`, to make it clearer that this is about the quantization loss (since it is a global variable in the configuration).

shadeMe (approved):

PR #25, opened 2022-10-12
Refactor pipeline and listeners to propagate all Transformer outputs

shadeMe:
Models that follow the transformer in the pipeline model chain operate on the outputs of all the transformer layers. This will allow for the usage of a separate listener that can pass a weighted sum of the former to the pooling layer (as opposed to using just the last layer's hidden state).

### DIFF
--- a/curated_transformers/models/__init__.py
+++ b/curated_transformers/models/__init__.py
@@ -1,4 +1,6 @@
 from .activations import GeluNew
+from .albert.encoder import AlbertEncoder
+from .bert.encoder import BertEncoder
 from .hf_wrapper import build_hf_encoder_loader_v1
 from .roberta.encoder import RobertaEncoder
 from .transformer_model import (
--- a/curated_transformers/models/hf_wrapper.py
+++ b/curated_transformers/models/hf_wrapper.py
@@ -1,19 +1,21 @@
-from typing import Tuple, List, Dict, Callable
+from typing import List
 from functools import partial
 
 from spacy.tokens import Doc
-from thinc.api import Model, torch2xp, get_current_ops
+from thinc.api import Model, get_current_ops
 from thinc.api import xp2torch, PyTorchWrapper_v2
 from thinc.shims.pytorch_grad_scaler import PyTorchGradScaler
-from thinc.types import ArgsKwargs, Floats2d, Ints1d, Ragged
+from thinc.types import ArgsKwargs, Floats2d, Ints1d
+from thinc.util import torch2xp
 
 from .._compat import transformers, has_hf_transformers
 from ..util import registry
-from .hf_util import convert_hf_pretrained_model_parameters
+from .hf_util import convert_hf_pretrained_model_parameters, SupportedHfTransformersT
+from .output import TransformerModelOutput
 
 
 def build_hf_transformer_encoder_v1(
-    encoder: Model[List[Doc], List[Ragged]],
+    encoder: SupportedHfTransformersT,
     *,
     mixed_precision: bool = False,
     grad_scaler_config: dict = {},
@@ -33,6 +35,9 @@ def build_hf_transformer_encoder_v1(
         grad_scaler=PyTorchGradScaler(**grad_scaler_config),
     )
 
+    # Actual value initialized by the parent Pipe instance.
+    model.attrs["_all_layer_outputs"] = True
+
     return model
 
 
@@ -48,7 +53,7 @@ def _convert_inputs(
     max_seq_len = max(x.size for x in X)
     if max_seq_len > max_model_seq_len:
         raise ValueError(
-            f"span window size of '{max_seq_len}' exceeds maximum allowed sequence length of '{max_model_seq_len}'"
+            f"Span window size of '{max_seq_len}' exceeds maximum allowed sequence length of '{max_model_seq_len}'"
         )
 
     # Transform the list of strided spans to a padded array.
@@ -60,7 +65,6 @@ def _convert_inputs(
     Xt = xp2torch(Xt)
 
     def convert_from_torch_backward(d_inputs):
-        # FIXME: d_inputs is an ellipsis here - bug?
         # No gradients for the inputs.
         return [ops.alloc1f(x.shape[0]) for x in X]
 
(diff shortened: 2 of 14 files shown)


curated_transformers/models/__init__.py
```diff
@@ -1,5 +1,11 @@
+from typing import Union
```
danieldk, line 1: Do we need this import here?

curated_transformers/models/hf_wrapper.py
```diff
@@ -72,24 +70,30 @@ def convert_from_torch_backward(d_inputs):
...
+    Yt = [
+        [output[i, :len, :] for output in model_outputs.all_outputs]
+        for i, len in enumerate(input_lens)
+    ]
+    Y = [[torch2xp(layer, ops=ops) for layer in output] for output in Yt]
+    output = TransformerModelOutput(outputs=Y)
+
+    def convert_for_torch_backward(dY: List[List[Floats2d]]):
+        Yt_flat = [y for inner in Yt for y in inner]
+        dYt_flat = [xp2torch(y) for inner in dY for y in inner]
```
danieldk, line 88: I think this already came up in an earlier discussion -- we'd always backprop through all layers, even if we only use last layer listeners. The performance impact may not be that bad, since the gradients are accumulated, but did you have a chance to measure the difference?

curated_transformers/models/output.py
```diff
@@ -26,9 +30,53 @@ def layer_hidden_state(self, idx: int) -> Tensor:
...
+    Also acts as the accumulator for the outputs of subsequent models in the Transformer pipeline."""
+
+    # Non-padded, un-stacked versions of the outputs.
+    # The outer list tracks Docs/spans and the inner
+    # list tracks the embedding + layer outputs of each Doc/span.
+    #
+    # The inner-most element is a Floats2d when returned by
+    # the PyTorchWrapper transformer model, which are subsequently
+    # converted to Ragged by the models that follow.
+    all_outputs: List[List[Union[Floats2d, Ragged]]]
```
danieldk, line 62: I wonder if we can make this a typevar with bounds rather than a union. Then we can reason about what is inside this container and e.g. `embedding_outputs` would return the correct type. Or is it actually a mixture. If not, and a typevar is not an option, the union should be the outermost type.

curated_transformers/models/with_strided_spans.py
```diff
@@ -1,16 +1,24 @@
-from typing import Callable, List, Tuple
+from typing import Callable, List, Tuple, Union
 from functools import partial
+from unittest.mock import sentinel
```
danieldk, line 3: Stray import?

curated_transformers/models/with_strided_spans.py
```diff
@@ -56,6 +66,14 @@ def with_strided_spans_forward(
 
     Y_layer, backprop_layer = model.layers[0](spans, is_train=is_train)
 
+    is_trf_output = isinstance(Y_layer, TransformerModelOutput)
```
danieldk, line 69: Could you explain this conditional block? In what case it the layer output not transformer output?

curated_transformers/models/with_strided_spans.py
```diff
@@ -1,16 +1,24 @@
...
+
 
 def build_with_strided_spans_v1(stride=96, window=128):
     return partial(with_strided_spans, stride=stride, window=window)
 
 
 def with_strided_spans(
     layer, *, stride=96, window=128
-) -> Model[List[Ragged], List[Ragged]]:
+) -> Model[List[Ragged], Union[TransformerModelOutput, RaggedInOutT]]:
```
danieldk, line 21: This ties in with the other comment. But I don't really understand why we need the unions here. Aren't we always dealing with transformer outputs?

curated_transformers/pipe.py
```diff
@@ -258,13 +262,15 @@ def predict(self, docs: Iterable[Doc]):
...
+    def set_annotations(
+        self, docs: Sequence[Doc], trf_output: TransformerModelOutput
+    ) -> None:
         """Modify a batch of documents, using pre-computed scores.
         docs (Iterable[Doc]): The documents to modify.
         tokvecses: The tensors to set, produced by Tok2Vec.predict.
         DOCS: https://spacy.io/api/tok2vec#set_annotations
         """
-        for doc, tokvecs in zip(docs, tokvecses):
+        for doc, tokvecs in zip(docs, trf_output.all_outputs):
```
danieldk, line 273: I wonder if we can do nicer here, by wrapping the per-doc outputs in another class. E.g. `DocTransformerModelOutput` (there is probably a better name). Then we don't need the `-1` indexing in other places, but could just have a property to access the last hidden layer.

curated_transformers/tokenization/sentencepiece_adapters.py
```diff
@@ -10,6 +12,8 @@
 _SPP_EOS = 2
 _SPP_UNK = 0
 
+RemoveBosEosInOutT = Union[List[Ragged], List[List[Ragged]], TransformerModelOutput]
```
danieldk, line 15: Same question about this type, don't we always have `TransformerModelOutput`?

curated_transformers/models/__init__.py
shadeMe: Seems like a leftover, will remove.

curated_transformers/models/hf_wrapper.py
shadeMe: Not yet, but I can train a model to check before we merge this PR.

curated_transformers/models/output.py
shadeMe: It's not a mixture in that all the innermost elements are of the same time at any given point, i.e., either a `Ragged` or a `Floats2d`. In fact, they are the former in all cases except inside the `with_strided_spans` model.

curated_transformers/models/with_strided_spans.py
shadeMe: Yup, will remove.

curated_transformers/models/with_strided_spans.py
shadeMe: At the moment, some of the tests for the `with_strided_spans` model simply pass in `Ragged` inputs. I thought it would be cleaner to support them instead of wrapping them in a `TransformerModelOutput` (and it makes the model usable outside of transformer contexts).

curated_transformers/models/with_strided_spans.py
shadeMe: As mentioned above, we support inner layers that either return a `TransformerModelOutput` or a (nested) list of `Ragged`s in the current implementation.

curated_transformers/pipe.py
shadeMe: Sounds good, I'll add that.

curated_transformers/tokenization/sentencepiece_adapters.py
shadeMe: Same answer as above.

curated_transformers/models/output.py
danieldk: So this could be `Union[List[List[Floats2d]], List[List[Ragged]]]` or even a type variable with bounds?

curated_transformers/models/with_strided_spans.py
danieldk: I think we should initially just make `with_strided_spans` work with `TransformerModelOutput`s. I think it's better to make the tests mock some `MakeModelOutputs` than to have union types and condition blocks to handle them this early on. It makes the types harder to read, the implementations more difficult to follow, and I think it is unlikely that these layers would be used outside `curated-transformers`.

curated_transformers/models/output.py
shadeMe: I've added a type variable for this.

curated_transformers/models/hf_wrapper.py
shadeMe: The performance impact of processing all layer outputs is significant (~20-23% reduction in WPS), but the latest changes should restrict the penalty to only those cases where all outputs are used downstream. When they aren't, we simply remove all but the last hidden representation from the `TransformerOutput` instance, thereby avoiding redundant processing.

curated_transformers/models/hf_wrapper.py
```diff
@@ -36,6 +35,10 @@ def build_hf_transformer_encoder_v1(
         grad_scaler=PyTorchGradScaler(**grad_scaler_config),
     )
 
+    # Managed by the parent Pipe instance. Remove all but the last
+    # layer outputs if the rest aren't required by the downstream models.
+    model.attrs["all_layer_outputs"] = True
```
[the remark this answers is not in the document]
shadeMe: After internal discussion, we've decided to implement this as a pipe-level, set-and-forget option.

curated_transformers/pipe.py
```diff
@@ -207,6 +225,14 @@ def add_listener(
         self, listener: "TransformerListener", component_name: str
     ) -> None:
         """Add a listener for a downstream component. Usually internals."""
+        if (
+            not isinstance(listener, LastTransformerLayerListener)
```
danieldk, line 229: I think this check should be in the listener, since there can be other custom listeners that only use the last layer.

shadeMe: Added `needs_all_layer_outputs` to `TransformerListener`.

danieldk (approved):

PR #26, opened 2022-10-17
Add support for RoBERTa models

danieldk:
This adds:

- `build_byte_bpe_encoder`: create an empty byte BPE encoder.
- `curated-transformers.HFByteBPELoader.v1`: load a RoBERTa byte BPE model from 🤗 transformers.
- `build_roberta_transformer_model_v1`: factory for RoBERTa models.


setup.cfg
```diff
@@ -22,6 +22,7 @@ spacy_factories =
 
 spacy_architectures =
     curated-transformers.BertTransformer.v1 = curated_transformers.models:build_bert_transformer_model_v1
+    curated-transformers.RobertaTransformer.v1 = curated_transformers.models:build_roberta_transformer_model_v1
```
shadeMe, line 25: Is there any particular reason why we don't need to/aren't adding the various piece encoder entry points to this section too?

shadeMe (approved):

PR #27, opened 2022-10-18
Add basic CI configuration

danieldk:
This adds a basic CI configuration for until we move this to explosion and setup Azure Devops.


.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Run tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.7', '3.10']
```
shadeMe, line 12: Might be a good idea to also test with 3.11rc2, like we're currently doing in our other repos.

danieldk: Maybe we can wait until 3.11 is released and bump it in another PR then? I want to use both the oldest and newest supported Python versions.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-version = 0.5.0
+version = 0.0.1
 description = Cutting-edge experimental spaCy components and features
```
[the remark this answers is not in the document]
danieldk: Ah, right, good catch!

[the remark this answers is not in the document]
danieldk: Fixed.

curated_transformers/tokenization/bbpe_encoder.py
```diff
@@ -0,0 +1,107 @@
+from typing import List, TypeVar
```
shadeMe, line 1:
```suggestion
from typing import List, Optional
```
Reverting unintended change (`TypeVar` is not used anywhere).

shadeMe:
Seems like the history got mangled in the last commit somehow? There are some errant commits from the Roberta PR (my new review comment above is apparently for a file modified in one of them).

danieldk:
> Seems like the history got mangled in the last commit somehow? There are some errant commits from the Roberta PR (my new review comment above is apparently for a file modified in one of them).

Not sure how that commit came back, I already dropped it sometime yesterday. Should be fixed now.

curated_transformers/tokenization/bbpe_encoder.py
danieldk: I think this came from the commit that shouldn't have been there. Fixed now.

shadeMe (approved):

PR #28, opened 2022-10-19
Merge all HF piece loader factories into a single factory

danieldk:
This makes loading piece encoders from Huggingface Hub similar to loading transformer encoders.

### DIFF
--- a/curated_transformers/tokenization/hf_loader.py
+++ b/curated_transformers/tokenization/hf_loader.py
@@ -1,0 +1,78 @@
+from typing import List
+import json
+from spacy.tokens import Doc
+from thinc.api import Model
+
+from .._compat import has_hf_transformers, transformers
+from ..util import registry
+from .bbpe_encoder import ByteBPEProcessor
+from .sentencepiece_encoder import SentencePieceProcessor
+from .wordpiece_encoder import WordPieceProcessor
+
+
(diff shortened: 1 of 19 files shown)


curated_transformers/tokenization/hf_loader.py
```diff
@@ -0,0 +1,76 @@
...
+
+def _convert_encoder(model: Model, tokenizer: "transformers.PreTrainedTokenizerBase"):
+    if isinstance(tokenizer, transformers.BertTokenizerFast):
+        return _convert_wordpiece_encoder(model, tokenizer)
+    elif isinstance(tokenizer, transformers.RobertaTokenizerFast):
+        return _convert_byte_bpe_encoder(model, tokenizer)
+    elif isinstance(tokenizer, transformers.XLMRobertaTokenizerFast):
+        return _convert_sentencepiece_encoder(model, tokenizer)
+
+    raise ValueError("Loading from this 🤗 tokenizer is not supported")
```
shadeMe, line 33:
```suggestion
    raise ValueError(f"Loading from the '{type(tokenizer}' 🤗 tokenizer is not supported")
```
We can add a bit more context to the error message here.

danieldk: Great idea, fixed!

PR #36, opened 2022-10-25
Add scalar weighting model and listener

shadeMe:
Field names in `PyTorchTransformerOutput`, `TransformerModelOutput` and `DocTransformerOutput` were additionally modified to reduce ambiguity.

### DIFF
--- a/curated_transformers/models/scalar_weight.py
+++ b/curated_transformers/models/scalar_weight.py
@@ -1,0 +1,125 @@
+from typing import List
+from thinc.layers.pytorchwrapper import PyTorchWrapper_v2
+from thinc.model import Model
+from thinc.types import ArgsKwargs, Ragged
+from thinc.shims.pytorch_grad_scaler import PyTorchGradScaler
+from thinc.util import torch2xp, xp2torch
+import torch
+from torch import Tensor
+from torch.nn import Module
+from spacy.util import SimpleFrozenDict
+
+from ..util import all_equal
+
+
+# From syntaxdot:
+# https://github.com/tensordot/syntaxdot/blob/22bd3d43ed2d7fcbef8a6217b01684194fae713f/syntaxdot-transformers/src/scalar_weighting.rs#L62
+class ScalarWeight(Module):
+    def __init__(self, *, num_layers: int, dropout_prob: float = 0.1):
+        super().__init__()
+
+        self.layer_weights = torch.nn.parameter.Parameter(torch.zeros(num_layers))
+        self.scale = torch.nn.parameter.Parameter(torch.tensor((1.0,)))
+        self.dropout_prob = dropout_prob
+
+    def forward(
+        self,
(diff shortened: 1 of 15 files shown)


shadeMe:
If we hold off on merging until [this `thinc` PR](https://github.com/explosion/thinc/pull/796) gets merged into `master`, we can pin the `thinc` requirement directly on the merge commit.

curated_transformers/models/scalar_weight.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, *, num_layers: int, dropout_prob: float = 0.1):
+        super().__init__()
+
+        self.layer_weights = torch.nn.parameter.Parameter(
+            torch.zeros((num_layers), requires_grad=True)
+        )
+        self.scale = torch.nn.parameter.Parameter(
+            torch.tensor((1.0,), requires_grad=True)
+        )
+        self.dropout = torch.nn.Dropout(p=dropout_prob)
```
danieldk, line 30: I think we can just store the probability, since the dropout layer itself is never used?

curated_transformers/pipe.py
```diff
@@ -175,7 +192,108 @@ def backprop(dYs):
...
                 outputs.append(model.ops.alloc2f(len(doc), width))
             else:
-                output, _ = pooling(doc._.trf_data.last_hidden_state, is_train)
+                output, _ = pooling(doc._.trf_data.last_hidden_layer_state, is_train)
+                outputs.append(output)
+
+        return outputs, lambda dX: []
+
+
+class ScalarWeightingLayerListener(TransformerListener):
```
danieldk, line 201: Maybe just `ScalarWeightingListener`?

curated_transformers/models/scalar_weight.py
```diff
@@ -0,0 +1,147 @@
...
+    ops = model.ops
+    layer_hidden_sizes = [x.data.shape[1] for x in X]
+    if not all_equal(layer_hidden_sizes):
+        raise ValueError(
+            f"Not all hidden sizes are equal in input passed to scalar weight"
+        )
+
+    Xt = []
+    for layer in X:
+        batch = ops.alloc3f(1, layer.data.shape[0], layer.data.shape[1])
```
danieldk, line 117: Is it necessary to have the initial singleton dimension? Maybe it's better to modify the PyTorch model to not require this dimension?

Another thing that I wondered about -- since we are copying here anyway, maybe we should allocate + fill a single tensor for all layers, so that the `stack` is not necessary in the PyTorch model?

shadeMe: I originally had it implemented in such a way that we'd pass the entire document batch as its input. But it turned out to be unwieldy as the pooling didn't support batching. I'll stack the tensors without the batch dim during input conversion.

danieldk (approved):

PR #37, opened 2022-10-25
Register `slow` `pytest` marker in `pyproject.toml`

shadeMe:
Fixes the following warning when executing tests:
```
PytestUnknownMarkWarning: Unknown pytest.mark.slow - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
@pytest.mark.slow
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
-requires = [
-    "setuptools",
-    "wheel",
-]
+requires = ["setuptools", "wheel"]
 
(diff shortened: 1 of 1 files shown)


danieldk:
I was annoyed by the same warnings, but fixed it by adding the marker to `conftest.py`, since the option is also added there.
#38

shadeMe:
Closed in favour of #38.

PR #39, opened 2022-10-26
Pass `mixed_precision` parameter to PyTorch transformer encoders

shadeMe:


curated_transformers/models/transformer_model.py
```diff
@@ -41,7 +41,8 @@ def build_albert_transformer_model_v1(
     num_hidden_layers: int = 12,
     padding_idx: int = 0,
     type_vocab_size: int = 2,
-    torchscript=False,
+    torchscript: bool = False,
+    mixed_precision: bool = False,
```
danieldk, line 45: We also need to pass through the grad scaler configuration.

shadeMe: Fixed.

danieldk (approved):

PR #47, opened 2022-11-02
Merge key, query, and value input projections

danieldk:
Use a single `(hidden_size, 3 * hidden_size)` linear layer for computing the key, query, and value representations. This brings our implementation closer to that of PyTorch, allowing us to use their fused kernels more easily.

For simplicity, the multi-head attention is also changed to take a single input representation for key, query, and value.

### DIFF
--- a/curated_transformers/models/bert/layer.py
+++ b/curated_transformers/models/bert/layer.py
@@ -23,9 +23,7 @@ class BertSelfAttention(Module):
 
         self.dims_per_head = self.model_dim // self.num_heads
         self.attention = ScaledDotProductAttention(dropout_prob=config.dropout_prob)
-        self.query = torch.nn.Linear(self.model_dim, self.model_dim)
-        self.key = torch.nn.Linear(self.model_dim, self.model_dim)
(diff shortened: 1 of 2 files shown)


curated_transformers/models/bert/layer.py
```diff
@@ -50,9 +48,7 @@ def _combine_heads(self, x: Tensor) -> Tensor:
             x.transpose(1, 2).contiguous().view(batch_size, seq_len, head * model_dim)
         )
 
-    def forward(
-        self, k: Tensor, q: Tensor, v: Tensor, attn_mask: Optional[Tensor] = None
-    ) -> Tensor:
+    def forward(self, x: Tensor, attn_mask: Optional[Tensor] = None) -> Tensor:
         """
         Shapes:
             k, q, v - (batch, seq_len, model_dim)
```
shadeMe, line 54:
```suggestion
            x - (batch, seq_len, model_dim)
```

danieldk: Fixed.

shadeMe (approved):

PR #48, opened 2022-11-03
Attention mask handling improvements

danieldk:
- Add the `AttentionMask` class, which stores a boolean mask and creates a logit mask on-demand. The logit mask is cached to avoid recomputing it for every layer.
- Make the attention mask mandatory in internal classes.
- Move attention mask check and view logic to
`ScaledDotProductAttention.forward`.
- Fix inconsistent naming as `mask`.

### DIFF
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -1,9 +1,36 @@
 from typing import Optional
+from dataclasses import dataclass
 import math
 import torch
 from torch import Tensor
 from torch.nn import Module
 
+
+@dataclass
+class AttentionMask:
+    bool_mask: Tensor
+    _logit_mask: Optional[Tensor] = None
+
+    def __post_init__(self):
+        if self.bool_mask.dtype != torch.bool:
+            raise ValueError(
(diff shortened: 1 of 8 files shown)


curated_transformers/models/attention.py
```diff
@@ -1,10 +1,37 @@
...
+            raise ValueError(
+                f"attention mask of dtype torch.bool expected, was {self.bool_mask.dtype}"
+            )
+
+    @property
+    def logit_mask(self) -> Tensor:
+        if self._logit_mask is None:
+            # The value is `torch.finfo(attn_scores.dype).min`. Unfortunately,
+            # we cannot use `torch.finfo` in TorchScript.
+            self._logit_mask = (1.0 - self.bool_mask.int()) * -3.4028234663852886e38
```
shadeMe, line 25: Can't we get the minimum value from the built-in `float` type (instead of using the constant directly)?

danieldk: `float` is double precision and we need single precision.

shadeMe (approved):

PR #49, opened 2022-11-03
Add `gradual_transformer_unfreezing` callback

shadeMe:
Soft-dependency on https://github.com/explosion/spaCy/pull/11739.


curated_transformers/util.py
```diff
@@ -43,3 +48,81 @@ def all_equal(iterable):
...
+
+
+def gradual_transformer_unfreezing_per_pipe(
+    nlp: Language,
+    callback_args: Dict[str, Any],
+    freeze_params: Dict[str, int],
+    last_unfreeze_step: int,
+):
+    current_step = callback_args["step"]
+    if current_step > last_unfreeze_step:
```
danieldk, line 60: `last_unfreeze_step` makes the code less readable. Is this optimization really needed?

If it is, I'd do the `max`-ing here, since it is really an implementation detail of the function.

curated_transformers/util.py
```diff
@@ -43,3 +48,81 @@ def all_equal(iterable):
...
+):
+    if current_step == 0:
+        # Always freeze at the start of training.
+        pipe.frozen = True
+    elif current_step < unfreeze_step:
+        assert (
+            pipe.frozen == True
+        ), "Transformer frozen state unexpectedly changed before unfreezing"
+    elif current_step == unfreeze_step:
+        pipe.frozen = False
```
danieldk, line 106:
```suggestion
    pipe.frozen = current_step < unfreeze_step
```

I don't think it's good to have the assertion here, maybe the user made a composed callback?

curated_transformers/util.py
shadeMe: Probably not, to be honest. I think we'll be fine if it's removed.

danieldk (approved):

danieldk:
Oh, there is a conflict now, probably from the mypy PR?

shadeMe:
The test failure appears to be due to a network issue. All tests pass locally; will go ahead and merge this.

PR #50, opened 2022-11-04
Split query, key, value differently for Torch compat

danieldk:
I messed up the chunking order in the previous PR. Fix it once again for Torch compat.


danieldk:
Trained a German BERT model with these changes. Results look good.

shadeMe (approved):

PR #54, opened 2022-11-07
Add missing type annotations

shadeMe:


curated_transformers/tokenization/bbpe_encoder.py
```diff
@@ -81,7 +83,12 @@ def byte_bpe_encoder_forward(model: Model, X: List[Doc], is_train: bool):
     return pieces, lambda dY: []
 
 
-def build_byte_bpe_encoder_loader_v1(*, vocab_path: Path, merges_path: Path):
+def build_byte_bpe_encoder_loader_v1(
+    *, vocab_path: Path, merges_path: Path
+) -> Callable[
+    [Model[List[Doc], List[Ragged]], Optional[List[Doc]], Any],
```
danieldk, line 89: I think it would be nice to have aliases for complex, repeatedly used types like `Model[List[Doc], List[Ragged]]`. Semantic descriptions are so much easies to read (e.g. this being something like `Tok2PiecesModel`.

danieldk (commented):
Thanks, it's really nice to have these types specified and corrected. But I thing the readability of the types would benefit from some aliases of repeatedly used complex types. See the example that I commented on.

danieldk (approved):

PR #55, opened 2022-11-07
Implement BERT layer using `torch.nn.TransformerEncoderLayer`

danieldk:


danieldk:
We have decided to not do this and wait until there is a stable API to use the fastpath.

PR #65, opened 2022-11-16
Move all Torch modules to a subpackage

danieldk:
We were mixing Torch modules and Thinc models. Move all Torch modules to a subpackage.


shadeMe (approved):
Looks good. I think it would be a good idea to rebase this PR `main` after merging the [docs PR](https://github.com/danieldk/curated-transformers/pull/67), given that the former renames and moves files around.

PR #67, opened 2022-11-16
Add doc strings to public entry points and their associated classes

shadeMe:
Also includes some minor renaming.

### DIFF
--- a/curated_transformers/models/bert/encoder.py
+++ b/curated_transformers/models/bert/encoder.py
@@ -40,11 +40,6 @@ class BertEncoder(Module):
         """
         Shapes:
             input_ids, token_type_ids - (batch, seq_len)
-
-        `attention_mask` indicates elements to attend to with `True` (and `False` otherwise)
-
-        Returns a tuple of consisting of a list of tensors from each Transformer
-        layer and the sum of the input and positional embeddings.
         """
         if attention_mask is None:
             attention_mask = self._create_attention_mask(input_ids)
--- a/curated_transformers/models/roberta/encoder.py
+++ b/curated_transformers/models/roberta/encoder.py
@@ -39,11 +39,6 @@ class RobertaEncoder(Module):
         """
         Shapes:
             input_ids, token_type_ids - (batch, seq_len)
-
-        `attn_mask` indicates elements to attend to with `1` (and `0` otherwise)
-
-        Returns a tuple of consisting of a list of tensors from each Transformer
-        layer and the sum of the input and positional embeddings.
         """
         if attention_mask is None:
             attention_mask = self._create_attention_mask(input_ids)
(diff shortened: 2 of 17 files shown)


curated_transformers/models/bert/encoder.py
```diff
@@ -41,10 +41,7 @@ def forward(
         Shapes:
             input_ids, token_type_ids - (batch, seq_len)
 
-        `attention_mask` indicates elements to attend to with `True` (and `False` otherwise)
-
-        Returns a tuple of consisting of a list of tensors from each Transformer
-        layer and the sum of the input and positional embeddings.
+        `attention_mask` indicates elements to attend to with `True` (and `False` otherwise).
```
danieldk, line 44: We should leave the true/false part out here, since we now have a dedicated class.

curated_transformers/models/roberta/encoder.py
```diff
@@ -40,10 +40,7 @@ def forward(
         Shapes:
             input_ids, token_type_ids - (batch, seq_len)
 
-        `attn_mask` indicates elements to attend to with `1` (and `0` otherwise)
-
-        Returns a tuple of consisting of a list of tensors from each Transformer
-        layer and the sum of the input and positional embeddings.
+        `attn_mask` indicates elements to attend to with `1` (and `0` otherwise).
```
danieldk, line 43: We should leave the 0/1 part out here, since we now have a dedicated class.

curated_transformers/models/transformer_model.py
```diff
@@ -70,6 +70,52 @@ def build_albert_transformer_model_v1(
     mixed_precision: bool = False,
     grad_scaler_config: dict = SimpleFrozenDict(),
 ) -> TransformerModelT:
+    """Construct an ALBERT transformer model.
+
+    This model must be separately initialized using
```
danieldk, line 75: I think this is worded too strongly. It is not necessary and in some cases desirable to not use a loader (e.g. train a model from scratch/distillation).

curated_transformers/pipe.py
```diff
@@ -206,16 +235,19 @@ def predict(self, docs: Iterable[Doc]) -> TransformerModelOutput:
         return self.model.predict(docs)
 
     def set_annotations(
-        self, docs: Sequence[Doc], trf_output: TransformerModelOutput
+        self, docs: Sequence[Doc], predictions: TransformerModelOutput
```
danieldk, line 238: I think I like the original name better. _predictions_ sound like things like class labels.

danieldk (commented):
It's great to have all this documentation! Some minor comments.

curated_transformers/models/transformer_model.py
shadeMe: I've removed that sentence.

curated_transformers/pipe.py
shadeMe: Reverted.

PR #70, opened 2022-12-01
update readme to point to explosion

svlandeg:


danieldk:
See #71 , also updates `setup.cfg`.

svlandeg:
But but but I was first 🥇

PR #72, opened 2022-12-01
Replace GitHub actions CI by Azure pipelines

danieldk:

PR #73, opened 2022-12-01
Add `mypy` to requirements

shadeMe:
Copied from spaCy/requirements.txt. Should fix the Azure CI failures in #72.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ cutlery>=0.0.3,<0.1.0
 spacy>=3.4.0,<3.5.0
 srsly
 torch>=1.12.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -2,6 +2,7 @@ cutlery>=0.0.3,<0.1.0
 spacy>=3.4.0,<3.5.0
 srsly
 torch>=1.12.0
+mypy>=0.990,<0.1000; platform_machine != "aarch64" and python_version >= "3.7"
```
danieldk, line 5: Do we need the ARM restriction?

adrianeboyd: The restriction is to avoid using the non-compiled version in thinc or spacy because it is too slow.

danieldk (approved):

PR #79, opened 2023-01-04
Increase GPU throughput during gradient accumulation

shadeMe:
## Description

Originally, we were updating the pipe's cumulative loss value for every layer being backpropagated into, for every listener. This resulted in significant overhead when using the `ScalarWeightingListener` as it led to a lot of consecutive kernel launches/teardowns and additional device-to-host copies.

Now, we merely accumulate the gradients from all listeners and calculate the loss once at the end of the accumulation phase. This improves the throughput of the `ScalarWeightingListener`'s backprop pass considerably, reducing spent GPU time spent from 11.6% to 3.6% (20% and 62% reduction in average and median ms/batch times respectively; I'm trying the nVidia method of up-selling the improvements :trollface:).

This PR also refactors the output and backprop calculation code into a separate helper method.

### Types of change

Refactor/Feature/Optimization

## Checklist

### DIFF
--- a/curated_transformers/pipe.py
+++ b/curated_transformers/pipe.py
@@ -298,74 +298,9 @@ class Transformer(TrainablePipe):
         # To ensure that the model's internal state is always consistent with the pipe's.
         self._set_model_all_layer_outputs(self.all_layer_outputs)
 
-        if self.frozen:
-            # Ensures that the inner torch model is executed in a `no_grad` context.
-            outputs = self.model.predict(docs)
-            bp_outputs = None
-            d_outputs = None
-        else:
-            outputs, bp_outputs = self.model.begin_update(docs)
-            d_outputs = [
-                [
-                    Ragged(self.model.ops.alloc_f(t2v.dataXd.shape), t2v.lengths)
-                    for t2v in doc_layers
-                ]
-                for doc_layers in outputs.all_outputs
-            ]
-
-        def accumulate_gradient(
-            one_d_outputs: List[List[Ragged]], outputs_to_backprop: Tuple[int, ...]
-        ) -> None:
-            """Accumulate transformer loss and gradient. This is passed as a callback
-            to all but the last listener. Only the last one does the backprop.
-
-            `outputs_to_backprop` is a tuple of indices indicating to which layers/outputs
-            the gradients are to be propagated.
-            """
-            nonlocal d_outputs
-            assert d_outputs is not None
-            assert losses is not None
-            for i in range(len(one_d_outputs)):
-                for j in outputs_to_backprop:
-                    d_outputs[i][j].data += one_d_outputs[i][j].data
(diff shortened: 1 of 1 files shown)


curated_transformers/pipe.py
```diff
@@ -428,6 +363,76 @@ def finish_update(self, sgd: Optimizer) -> None:
...
+            for i in range(len(one_d_outputs)):
+                for j in outputs_to_backprop:
+                    d_outputs[i][j].data += one_d_outputs[i][j].data
+
+        def update_loss():
+            """Reduce the gradient buffer and update the losses dict."""
+            nonlocal cum_loss
+            for i in range(len(d_outputs)):
+                for j in range(len(d_outputs[i])):
+                    cum_loss += (d_outputs[i][j].data ** 2).sum()
```
danieldk, line 409: I wonder if we even need this. Since the loss of the transformer is just a rough indicator of the gradients of the loss of downstream components, I think we have some freedom in defining the loss. There are two straightforward definitions that I can think of:

1. The sum of an aggregate of the component gradients (current).
2. An aggregate of the accumulated gradient.

(2) would have the benefit that we only have to compute the squared sum of the gradients once. The losses might probably be slightly different, but sort of giving the same information when inspecting training progress: how large were the gradients of the downstream components in this step compared to the previous steps?

I can only think of one situation where (1) and (2) would be contradictory: suppose that we have two downstream components, and they have exactly the opposite gradients. Then (2) would be zero, whereas (1) would be some loss value. I am not sure which one I'd like more, (1) is more faithful from the components' perspective, but (2) more as an actual transformer loss, since in this case there is no update to be made and the transformer's parameters would be optimal with respect to the downstream components (moving the gradients in the direction of any downstream model would deteriorate the other model).

Though I think that concern is mostly theoretical.

shadeMe: Yeah, I too had a similar thought whilst I was refactoring the code. On the other hand, the practical issue is that we currently store the gradient buffer as a nested list of `Ragged` arrays. 

If I understand option (2) correctly, we'd need use a single, large buffer as the accumulator. This will require padding it to the largest document in the batch, maintaining the lengths of each incoming `Ragged` gradient array, and performing the requisite packing/unpacking when passing the accumulated buffer to the transformer model's backprop pass (since it expects a nested list of `Ragged`s). 

So, the overhead from the allocation/copying will probably eclipse the reduction in overhead from computing the square just once.

danieldk: Oh, wait, I think the code is now actually doing my proposed (2). To get the behavior from before this PR, you would actually have to apply `update_loss` in `accumulate_gradient` and not `backprop` (and compute the component loss from `one_d_outputs`. So this PR changed the semantics from (1) to (2), so it is not a refactor, but changes how the loss is computed.

Maybe there should then be a comment in the loss computation to state that the loss that is computed is different than that of other tok2vec implementations. (And also in future documentation.)

danieldk (approved):

PR #81, opened 2023-01-09
`ScalarWeightListener`: Batch documents in forward and backward passes

shadeMe:
## Description

Similar to #77, this PR concatenates the layer outputs of all documents into a batched tensor before performing the scalar weighting on it. This speeds up the corresponding listeners to a small extent (50% reduction in median ms/batch though /nvidia-talk) during both training and distillation.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/models/types.py
+++ b/curated_transformers/models/types.py
@@ -9,11 +9,11 @@ from .output import TransformerModelOutput
 PoolingModelT = Model[Ragged, Floats2d]
 
 WithRaggedLayersInT = Union[Iterable[Doc], Iterable[Iterable[Ragged]]]
-WithRaggedLayersOutT = List[List[Floats2d]]
+WithRaggedLayersOutT = List[List[Floats2d]]  # Doc -> Layer -> Representation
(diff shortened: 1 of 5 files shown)


curated_transformers/models/types.py
```diff
@@ -48,6 +48,6 @@
 TransformerBackpropT = Callable[[List[List[Floats2d]]], Any]
 TransformerModelT = Model[TransformerInT, TransformerOutT]
 
-ScalarWeightInT = List[Ragged]
-ScalarWeightOutT = Ragged
+ScalarWeightInT = List[List[Ragged]]
```
danieldk, line 51: We didn't really do this for other types, but maybe we should start adding small comments what the lists are here? 

Not sure what good notation would be?

I just want to make sure that if someone reads the types knows what the input to e.g. scalar weighting actually means. (docs -> layers -> representations vs. layers -> docs -> representations).

shadeMe: Good idea. `something -> something else` looks decently readable to me.

danieldk (approved):

PR #82, opened 2023-01-11
Add curated-transformers.TransformerLayersListener.v1

danieldk:
## Description

TransformerLayersListener passes through the pooled representations of all layers. This is useful for pipes that need access to all layers, such as layer distillation.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/listener.py
+++ b/curated_transformers/listener.py
@@ -21,7 +21,7 @@ from .models.types import (
     ScalarWeightOutT,
     ScalarWeightModelT,
 )
-from .models.output import DocTransformerOutput, TransformerModelOutput
+from .models.output import TransformerModelOutput
 from .models.pooling import with_ragged_layers, with_ragged_last_layer
 from .models.types import (
     WithRaggedLayersModelT,
@@ -196,7 +196,7 @@ class TransformerLayersListener(TransformerListener):
     its model.
     """
 
-    name = "transformer_layer_listener"
+    name = "transformer_layers_listener"
 
     def __init__(
         self,
@@ -287,7 +287,7 @@ def tranformer_layers_listener_forward(
         if any(no_trf_data):
             assert all(no_trf_data)
             return [
-                n_layers * [model.ops.alloc2f(len(doc), width)] for doc in docs
+                [model.ops.alloc2f(len(doc), width) for _ in range(n_layers)] for doc in docs
             ], lambda dY: []
 
         if any(doc._.trf_data.last_layer_only for doc in docs):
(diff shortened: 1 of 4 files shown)


curated_transformers/listener.py
```diff
@@ -21,6 +21,51 @@
     ScalarWeightOutT,
     ScalarWeightModelT,
 )
+from .models.output import DocTransformerOutput, TransformerModelOutput
```
shadeMe, line 24:
```suggestion
from .models.output import TransformerModelOutput
```
Unused import.

curated_transformers/listener.py
```diff
@@ -144,6 +189,113 @@ def verify_inputs(self, inputs: Iterable[Doc]) -> bool:
...
 
 
+class TransformerLayersListener(TransformerListener):
+    """Passes through the pooled representations of all transformer layers.
+
+    Requires its upstream Transformer component to return all layer outputs from
+    its model.
+    """
+
+    name = "transformer_layer_listener"
```
shadeMe, line 199:
```suggestion
    name = "transformer_layers_listener"
```

curated_transformers/listener.py
```diff
@@ -144,6 +189,113 @@ def verify_inputs(self, inputs: Iterable[Doc]) -> bool:
...
+            self,
+            name=self.name,
+            forward=tranformer_layers_listener_forward,
+            dims={"nO": width},
+            layers=[with_ragged_layers(pooling)],
+            attrs={
+                "grad_factor": grad_factor,
+                "layers": layers + 1,
+            },
+            refs={"pooling": pooling},
```
shadeMe, line 236: Do we refer to this layer elsewhere in the codebase?

curated_transformers/listener.py
```diff
@@ -144,6 +189,113 @@ def verify_inputs(self, inputs: Iterable[Doc]) -> bool:
...
+        return Y, backprop
+    else:
+        width = model.get_dim("nO")
+
+        no_trf_data = [doc._.trf_data is None for doc in docs]
+        if any(no_trf_data):
+            assert all(no_trf_data)
+            return [
+                n_layers * [model.ops.alloc2f(len(doc), width)] for doc in docs
+            ], lambda dY: []
```
shadeMe, line 291: This will create a list with the same array in each index, I think? 
```suggestion
            return [
                [model.ops.alloc2f(len(doc), width) for _ in range(n_layers)] for doc in docs
            ], lambda dY: []
```

curated_transformers/listener.py
danieldk: Yeah, the distiller needs access to the pooler to apply it to the teacher outputs.

shadeMe (approved):

PR #83, opened 2023-01-13
Distillation: Add `mean` and `squared_l2_norm` normalization for hidden layers

shadeMe:
## Description

Port the SyntaxDot `squared_l2_norm` [normalization scheme](https://github.com/tensordot/syntaxdot/pull/227) and expose it to the config.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/pipeline/distill.py
+++ b/curated_transformers/pipeline/distill.py
@@ -8,6 +8,7 @@ from spacy.training import validate_get_examples
 from thinc.api import Config, Model, Ops, Optimizer, Ragged, set_dropout_rate
 
 from ..models.pooling import with_ragged_layers
+from ..models.distill import MSELoss
 
 DEFAULT_CONFIG_STR = """
     [transformer_distiller]
@@ -68,13 +69,7 @@ class TransformerDistiller(TrainablePipe):
         self.model = model
         self.name = name
         self.layer_mapping = None
-
-        expected_loss_normalization = ("mean", "squared_l2_norm")
-        if loss_normalization not in expected_loss_normalization:
-            raise ValueError(
-                f"Loss normalization must one of the following values: {expected_loss_normalization}"
-            )
-        self.loss_normalization = loss_normalization
+        self.mse_loss = MSELoss(self.model.ops, normalization=loss_normalization)
 
     def _layer_mapping(self, teacher_hidden, student_hidden):
         if self.layer_mapping is not None:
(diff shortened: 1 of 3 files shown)


curated_transformers/pipeline/distill.py
```diff
@@ -118,37 +127,49 @@ def distill(
...
-
-        losses[self.name] += sum(
-            sum(float((d_layer**2).sum()) for d_layer in d) for d in d_mse
-        )
-
-        student_backprop(d_mse)
-        if sgd not in (None, False):
+        # Transpose the representations to the get the following shape: [layer, batch, seq, repr]
+        t_student_hidden = list(zip(*student_hidden))
+        t_teacher_hidden = list(zip(*teacher_hidden))
```
danieldk, line 137: Nitpicking, but I think it is common to add a suffix for transpositions.

```suggestion
        student_hidden_t = list(zip(*student_hidden))
        teacher_hidden_t = list(zip(*teacher_hidden))
```

curated_transformers/pipeline/distill.py
```diff
@@ -118,37 +127,49 @@ def distill(
...
+
+        # Allocate loss accumulators on the GPU if possible.
+        cum_loss = self.model.ops.alloc1f(1)
+        layer_loss = self.model.ops.alloc1f(1)
+        t_mses = []
+        for student_layer_idx, teacher_layer_idx in layer_mapping:
+            student_batches = t_student_hidden[student_layer_idx]
+            teacher_batches = t_teacher_hidden[teacher_layer_idx]
```
danieldk, line 145:
```suggestion
            student_batch = t_student_hidden[student_layer_idx]
            teacher_batch = t_teacher_hidden[teacher_layer_idx]
```

I think? Now that the layer is indexed, this is just a batch.

curated_transformers/pipeline/distill.py
```diff
@@ -118,37 +127,49 @@ def distill(
...
+                    layer_loss += mse.sum()
+                elif self.loss_normalization == "squared_l2_norm":
+                    norm = self.model.ops.xp.linalg.norm(t_doc.reshape(-1)) ** 2
+                    mse /= norm
+                    layer_loss += mse.sum()
+                batch_mses.append(mse)
+
+            t_mses.append(batch_mses)
+            if self.loss_normalization == "mean":
+                cum_loss += layer_loss / (batch_size * n_elements)
```
danieldk, line 164: I have to recheck the derivatives, but I with mean loss, the gradients should also be scaled. Empirical example:

```
>>> X = torch.full((10,), 0.5, requires_grad=True)
>>> loss = ((X - torch.ones(10)) ** 2).mean()
>>> loss.backward()
>>> X.grad
tensor([-0.1000, -0.1000, -0.1000, -0.1000, -0.1000, -0.1000, -0.1000, -0.1000,
        -0.1000, -0.1000])
```

The code above would result in a gradient of `-0.5` for all parameters.

Maybe it's worth lifting the loss/gradient calculation out of this loop, to that we can test against Torch in a unit test? To easy to get it wrong...

curated_transformers/pipeline/distill.py
shadeMe: Fixed.

curated_transformers/pipeline/distill.py
shadeMe: Inlined the indexing op at the callsite instead.

curated_transformers/pipeline/distill.py
shadeMe: I've refactored the loss calculation into its own class and added unit tests.

PR #84, opened 2023-01-20
Distillation: Make `TransformerDistiller` a no-op in regular pipelines

shadeMe:
## Description

Override and stub `TrainablePipe` methods that would be called in a regular training/inference pipeline.

### Types of change

Fix

## Checklist

### DIFF
--- a/curated_transformers/pipeline/distill.py
+++ b/curated_transformers/pipeline/distill.py
@@ -172,27 +172,8 @@ class TransformerDistiller(TrainablePipe):
     def predict(self, docs: Iterable[Doc]) -> Any:
         return None
 
-    def set_annotations(self, docs: Iterable[Doc], scores) -> None:
+    def set_annotations(self, docs: Iterable[Doc], scores: Optional[Any]) -> None:
         pass
-
-    def update(
-        self,
-        examples: Iterable["Example"],
-        *,
-        drop: float = 0.0,
-        sgd: Optional[Optimizer] = None,
-        losses: Optional[Dict[str, float]] = None,
-    ) -> Dict[str, float]:
-        if losses is None:
-            losses = {}
-        losses.setdefault(self.name, 0.0)
-        return losses
-
-    def finish_update(self, sgd: Optimizer) -> None:
-        pass
-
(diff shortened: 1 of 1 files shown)


curated_transformers/pipeline/distill.py
```diff
@@ -168,5 +169,31 @@ def initialize(self, get_examples, *, nlp=None):
         assert len(doc_sample) > 0, Errors.E923.format(name=self.name)
         self.model.initialize(X=doc_sample)
 
-    def pipe(self, stream: Iterable[Doc], *, batch_size: int = 128) -> Iterator[Doc]:
-        return iter(stream)
+    def predict(self, docs: Iterable[Doc]) -> Any:
+        return None
+
+    def set_annotations(self, docs: Iterable[Doc], scores) -> None:
```
danieldk, line 175: I think `scores` can be `NoneType`?

curated_transformers/pipeline/distill.py
```diff
@@ -168,5 +169,31 @@ def initialize(self, get_examples, *, nlp=None):
...
+    ) -> Dict[str, float]:
+        if losses is None:
+            losses = {}
+        losses.setdefault(self.name, 0.0)
+        return losses
+
+    def finish_update(self, sgd: Optimizer) -> None:
+        pass
+
+    def add_label(self, label: str) -> int:
```
danieldk, line 194: We should probably not override this one, so that we get an error when someone tries to override this.

shadeMe: Just `add_label`?

danieldk: Good question! Maybe `update` too, because a user (like `Language`) should just not call this method when `is_trainable == False`.

shadeMe:
The test failure appears to be from the target branch.

danieldk:
> The test failure appears to be coming from the target branch.

We probably need to bump it to a newer spaCy version:

```
ImportError: cannot import name 'validate_distillation_examples' from 'spacy.training'
```

danieldk (approved):

PR #85, opened 2023-02-02
Batch the spans that are passed to the torch transformer model

shadeMe:
## Description

This places an upper-bound on the actual batch size of the transformer inputs. Previously, a single, very long document could yield a very large batch of inputs, potentially resulting in an OOM situation.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/tests/models/test_with_strided_spans.py
+++ b/curated_transformers/tests/models/test_with_strided_spans.py
@@ -123,12 +123,17 @@ def test_with_strided_spans_averaging():
     ops.xp.testing.assert_array_equal(dX[0].lengths, lengths)
 
 
-def test_incorrect_strides_and_batch_sizes_are_rejected():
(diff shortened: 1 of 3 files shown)


curated_transformers/tests/models/test_with_strided_spans.py
```diff
@@ -114,9 +123,13 @@ def test_with_strided_spans_averaging():
     ops.xp.testing.assert_array_equal(dX[0].lengths, lengths)
 
 
-def test_incorrect_strides_are_rejected():
+def test_incorrect_strides_and_batch_sizes_are_rejected():
```
danieldk, line 126: Minor nitpick, I'd make this two separate tests.

danieldk (commented):
Small nitpick.

Also, we should consider if we still need the `max_batch_items` option with this change?

danieldk (approved):

PR #88, opened 2023-02-06
Bump version to 0.0.2

danieldk:
## Description

Also expose the version through a Python module like Thinc/spaCy.

### Types of change

Maintenance

## Checklist

### DIFF
--- a/curated_transformers/about.py
+++ b/curated_transformers/about.py
@@ -1,0 +1,2 @@
+__version__ = "0.0.2"
+__release__ = True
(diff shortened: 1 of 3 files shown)


adrianeboyd:
`about.py` pre-dates better alternatives, I would just put the version in `setup.cfg` instead.

adrianeboyd:
You can get the version with `importlib.metadata.version(name)`. (I would try to transition all our packages away from `about.py` in their next major release.)

danieldk:
> You can get the version with `importlib.metadata.version(name)`. (I would try to transition all our packages away from `about.py` in their next major release.)

Ah, then I'll revert those changes. I know that you can use importlib, but it always seems a bit unwieldy.

PR #94, opened 2023-02-07
Prefix encoders with `curated_encoder`

danieldk:
## Description

This makes it easier to find parameters specific to curated transformers. This could e.g. be used for discriminative learning rates.

## Checklist

### DIFF
--- a/curated_transformers/models/pytorch/curated_transformer.py
+++ b/curated_transformers/models/pytorch/curated_transformer.py
@@ -31,4 +31,8 @@ class CuratedTransformer(Generic[CuratedEncoderT], Module):
         attention_mask: Optional[AttentionMask] = None,
         token_type_ids: Optional[Tensor] = None,
     ) -> PyTorchTransformerOutput:
+        """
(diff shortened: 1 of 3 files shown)


curated_transformers/models/pytorch/curated_transformer.py
```diff
@@ -0,0 +1,34 @@
...
+        super().__init__()
+        self.curated_encoder = encoder
+
+    def forward(
+        self,
+        input_ids: Tensor,
+        attention_mask: Optional[AttentionMask] = None,
+        token_type_ids: Optional[Tensor] = None,
+    ) -> PyTorchTransformerOutput:
+        return self.curated_encoder.forward(input_ids, attention_mask, token_type_ids)
```
shadeMe, line 34: Let's also document the expected shapes of the parameters in the docstring like we do with the other modules.

danieldk: Done.

shadeMe (approved):

PR #105, opened 2023-02-10
Add `build_bert_wordpiece_encoder`

danieldk:
## Description

Huggingface BERT models preprocess strings by tokenizing them on punctuation characters. Since Huggingface tokenization has become the defacto tokenization for BERT models, we should provide the same type of tokenization to be compatible.

We expose the piecer with punctuation tokenization using a new `build_bert_wordpiece_encoder`, keeping `build_wordpiece_encoder` as a vanilla encoder.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/curated_transformers/tokenization/__init__.py
+++ b/curated_transformers/tokenization/__init__.py
@@ -1,4 +1,7 @@
 from .bbpe_encoder import build_byte_bpe_encoder_loader_v1
 from .hf_loader import build_hf_piece_encoder_loader_v1
 from .sentencepiece_encoder import build_sentencepiece_encoder_loader_v1
-from .wordpiece_encoder import build_wordpiece_encoder_loader_v1
+from .wordpiece_encoder import (
+    build_bert_wordpiece_encoder,
+    build_wordpiece_encoder_loader_v1,
+)
--- a/curated_transformers/tokenization/wordpiece_encoder.py
+++ b/curated_transformers/tokenization/wordpiece_encoder.py
@@ -1,5 +1,6 @@
 from typing import Callable, Optional, Tuple
 from pathlib import Path
+import unicodedata
 
 from cutlery import WordPieceProcessor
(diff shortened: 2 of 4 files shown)


curated_transformers/tokenization/__init__.py
```diff
@@ -1,4 +1,7 @@
 from .bbpe_encoder import build_byte_bpe_encoder_loader_v1
 from .hf_loader import build_hf_piece_encoder_loader_v1
 from .sentencepiece_encoder import build_sentencepiece_encoder_loader_v1
-from .wordpiece_encoder import build_wordpiece_encoder_loader_v1
+from .wordpiece_encoder import (
+    build_bert_wordpiece_encoder,
```
shadeMe, line 5: Any reason why we're including this function and not any of the other (`build_xxxpiece_encoder` functions)?

curated_transformers/tokenization/wordpiece_encoder.py
```diff
@@ -26,6 +27,27 @@ def deserialize_my_custom_class(
     return WordPieceProcessor(value.decode("utf8").split("\n"))
 
 
+def build_bert_wordpiece_encoder() -> Tok2PiecesModelT:
+    """Construct a WordPiece piece encoder model that accepts a list
```
shadeMe, line 31: Let's add a short note explaining why we need a separate builder for BERT models in this docstring.

curated_transformers/tokenization/__init__.py
danieldk: Oh, good question. We should probably start exporting them, it's weird to export the loaders but not the constructors. Maybe in a separate PR?

shadeMe: Sounds good!

curated_transformers/tokenization/wordpiece_encoder.py
danieldk: 👍 done

shadeMe (approved):

PR #110, opened 2023-02-15
Add character encoder

danieldk:
## Description

This encoder splits a token into characters and looks up their identifiers. The primary motivation for adding this encoder is that our standard model for Japanese requires it. Along with the encoder itself, `curated-transformers.CharEncoderLoader.v1` is added to load character vocabularies and support to load Japanese character-based encoders is added to `curated-transformers.HFPieceEncoderLoader.v1`.

### Types of change

Feature

## Checklist


adrianeboyd:
Nothing jumps out at me, but I didn't try to double-check the token IDs in the tests or anything. I was surprised that the vocab file is just a plain `.txt` list, but I guess that's the typical format?

danieldk:
> I was surprised that the vocab file is just a plain `.txt` list, but I guess that's the typical format?

Yeah, luckily character vocabs and wordpiece vocabs are plain text.

polm (approved):
The fugashi/Sudachi parts look good to me! I'm not as familiar with the HF parts, but nothing looked off there either.

danieldk:
> "luckily" is not how I'd describe it.

Betty than some whacky JSON format, re-encoding UTF-8 pieces in its own encoding 😱 .

PR #111, opened 2023-02-16
Add `debug pieces` subcommand

danieldk:
This subcommand prints piece statistics for a given model and train/dev corpora.

Example output:

```
================== Training corpus statistics ===================
Median token length: 1.0
Mean token length: 1.34
Token length range: [1, 24]

================== Development corpus statistics ================
Median token length: 1.0
Mean token length: 1.32
Token length range: [1, 24]
```

## Checklist

- [ ] I confirm that I have the right to submit this contribution under the project's MIT license.

### DIFF
--- a/curated_transformers/cli/debug_pieces.py
+++ b/curated_transformers/cli/debug_pieces.py
@@ -1,0 +1,111 @@
+from typing import Any, Dict, List, Optional, cast
+import numpy
+from pathlib import Path
+from spacy import util
+from spacy.cli._util import (
+    debug_cli,
+    import_code,
+    parse_config_overrides,
+    show_validation_error,
+)
+from spacy.schemas import ConfigSchemaTraining
+from spacy.tokens import Doc
+from spacy.util import registry, resolve_dot_names
+from typer import Argument as Arg, Context, Option as Opt
+from wasabi import Printer, msg
+
+from ..tokenization.types import Tok2PiecesModelT
+from ..pipeline.transformer import Transformer
+
+
+@debug_cli.command(
+    "pieces",
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def debug_pieces_cli(
+    # fmt: off
+    ctx: Context,  # This is only used to read additional arguments
+    config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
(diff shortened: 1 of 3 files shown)


curated_transformers/cli/debug_pieces.py
```diff
@@ -0,0 +1,111 @@
...
+from spacy.cli._util import (
+    debug_cli,
+    import_code,
+    parse_config_overrides,
+    show_validation_error,
+)
+from spacy.schemas import ConfigSchemaTraining
+from spacy.tokens import Doc
+from spacy.util import registry, resolve_dot_names
+from typer import Argument as Arg, Context, Option as Opt
```
svlandeg, line 14: Note that we'll be [moving](https://github.com/explosion/spaCy/pull/12258) from `typer` to `radicli` for `v4` so maybe it makes sense already using that here now?

curated_transformers/cli/debug_pieces.py
```diff
@@ -0,0 +1,111 @@
...
+    # fmt: off
+    ctx: Context,  # This is only used to read additional arguments
+    config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
+    code_path: Optional[Path] = Opt(None, "--code-path", "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    transformer_name: Optional[str] = Opt(None, "--name", "-n", help="Name of the transformer pipe to gather piece statistics for (default: first transformer pipe)."),
+    # fmt: on
+):
+    """
+    Analyze word- or sentencepiece statistics.
+    DOCS: https://spacy.io/api/cli#debug-data
```
svlandeg, line 35:
```suggestion
    DOCS: https://spacy.io/api/cli#debug-pieces
```
(and TODO: actually write those docs :-) )

curated_transformers/cli/debug_pieces.py
danieldk: But this needs to be compatible with spaCy 3.6, would combining these work? We plan to make a `v4` branch of curated transformers for distillation. Maybe we can change it there later?

svlandeg: It probably wouldn't work if spaCy 3.6 were used with an older Python version, so yea, let's keep this for the `v4` branch.

PR #113, opened 2023-02-16
Add optional `piece_encoder` argument to each transformer architecture

danieldk:
## Description

This allows a configuration to override the piece encoder that is used by the architecture, which is needed in rare cases. For instance, the `cl-tohoku/bert-base-japanese-char-v2` model uses a BERT transformer but a character-based piecer.

By default the `piece_encoder` value is `None`, which will use the default piece encoder for the architecture. This ensures that there is no configuration overhead in the vast majority of cases.

The PR also adds the piece encoders to the registry and tests whether they can be fetched from the registry in a specific test in `test_registry.py`.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/models/architectures.py
+++ b/curated_transformers/models/architectures.py
@@ -26,16 +26,12 @@ from .pytorch.hf_util import convert_pretrained_model_for_encoder
 from .pytorch.roberta import RobertaConfig, RobertaEncoder
 from .remove_eos_bos import remove_bos_eos
 from .with_non_ws_tokens import with_non_ws_tokens
-from ..tokenization.bbpe_encoder import build_byte_bpe_encoder
-from ..tokenization.sentencepiece_encoder import (
-    build_camembert_sentencepiece_encoder,
-    build_sentencepiece_encoder,
-    build_xlmr_sentencepiece_encoder,
-)
-from ..tokenization.sentencepiece_encoder import build_sentencepiece_encoder
-from ..tokenization.wordpiece_encoder import (
-    build_bert_wordpiece_encoder,
-    build_wordpiece_encoder,
+from ..tokenization.bbpe_encoder import build_byte_bpe_encoder_v1
+from ..tokenization import (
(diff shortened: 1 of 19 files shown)


vin-ivar (approved):

curated_transformers/models/architectures.py
```diff
@@ -148,7 +147,9 @@ def build_albert_transformer_model_v1(
             grad_scaler_config=grad_scaler_config,
         )
 
-    piece_encoder = build_sentencepiece_encoder()
+    piece_encoder = (
+        build_sentencepiece_encoder_v1() if piece_encoder is None else piece_encoder
+    )
```
svlandeg, line 152: Can't we have the default `v1` in the config as default value instead of `None`? Then we wouldn't have this hidden default here.

svlandeg: (same comment for other places in this PR)

curated_transformers/tests/tokenization/test_registry.py
```diff
@@ -0,0 +1,36 @@
...
+        "curated-transformers.CharEncoderLoader.v1",
+        "curated-transformers.HFEncoderLoader.v1",
+        "curated-transformers.HFPieceEncoderLoader.v1",
+        "curated-transformers.PyTorchCheckpointLoader.v1",
+        "curated-transformers.SentencepieceLoader.v1",
+        "curated-transformers.WordpieceLoader.v1",
+    ],
+)
+def test_encoder_loader_from_registry(loader_name):
+    # Can't be contstructed, since most loaders have mandatory arguments.
```
svlandeg, line 35:
```suggestion
    # Can't be constructed, since most loaders have mandatory arguments.
```

curated_transformers/models/architectures.py
danieldk: Fixed.

svlandeg (approved):
Looks good :-)

PR #117, opened 2023-03-07
Install package from sdist in CI

shadeMe:
## Description

Updates the CI steps to use our standard workflow of installing packages from sdist archives.

### Types of change

Chore

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,19 +31,45 @@ jobs:
           cache: "pip" # caching pip dependencies
 
       - name: Install requirements
-        run: pip install -r requirements.txt
+        run: |
+          pip install --upgrade pip setuptools wheel
+          pip install -r requirements.txt
 
       - name: Install optional dependencies for Japanese models
         run: pip install fugashi[unidic-lite] spacy[ja]
 
-      - name: Install package
-        run: pip install .
+      - name: Build sdist
+        run: |
+          python setup.py build_ext --inplace
+          python setup.py sdist --formats=gztar
 
       - name: Run mypy
         run: mypy curated_transformers
 
+      - name: Delete source directory
+        run: |
+          rm -rf curated_transformers
+        shell: bash
+
+      - name: Uninstall all packages
+        run: |
+          pip freeze
+          pip freeze --exclude pywin32 > installed.txt
+          pip uninstall -y -r installed.txt
+
+      - name: Install from sdist
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -31,19 +31,46 @@ jobs:
...
 
       - name: Install optional dependencies for Japanese models
         run: pip install fugashi[unidic-lite] spacy[ja]
 
-      - name: Install package
-        run: pip install .
+      - name: Build sdist
+        run: |
+          python setup.py build_ext --inplace
+          python setup.py sdist --formats=gztar
```
adrianeboyd, line 44: Using `python setup.py` is discouraged/deprecated, so it's better to switch to a newer alternative. I would use `python -m build --sdist` instead (you'd have to install it above, too).

adrianeboyd: (I would really like to have a few actions or templates whatever to do a few of these small sets of steps that are done in every repo, but we haven't figured out how to do it in a way that it doesn't either look like a public action that others could use or a private action that users then couldn't actually inspect.)

shadeMe: `build` doesn't appear to have a separate action liker `build_ext` to build Cython extensions, though. On the other hand, do we actually need to invoke that action in the CI to begin with? It seems unnecessary given that  the Cython source files will be included in the sdist archive in either case, and the subsequent `pip install dist/$SDIST` call will build and install the extension artifacts to the correct folder in the active Python environment.

adrianeboyd: You don't need to run `build_ext` for this step. `python -m build --sdist` works fine. I've only updated some repos so far, but see: https://github.com/explosion/spaCy/blob/260cb9c6febc5c7c2c17686d4f4e62c13b6833b9/.github/azure-steps.yml#L17-L24

adrianeboyd: The compiled extensions are never in the sdist and we actually exclude the generated cython c/cpp files in most packages. (This is a more difficult decision, but we've had more problems with generated cython code being invalid for future python versions than with differences in the generated c code leading to differences. Cython officially recommends packaging the generated files and they're only excluded manually by us in `MANIFEST.in`.)

adrianeboyd: Parts of `setup.py` are still run by `python -m build`, what is deprecated is calling `setup.py` as a standalone script.

shadeMe: Thanks for the context! I've updated the step as suggested.

PR #119, opened 2023-03-08
Rename `HFEncoderLoader.v1` to `HFTransformerEncoderLoader.v1`

shadeMe:
## Description

To avoid ambiguitiy as we already have a `HFPieceEncoderLoader` entry point.

### Types of change

Refactor

## Checklist


setup.cfg
```diff
@@ -51,7 +51,7 @@ spacy_cli =
 thinc_model_loaders =
     curated-transformers.ByteBPELoader.v1 = curated_transformers.tokenization:build_byte_bpe_encoder_loader_v1
     curated-transformers.CharEncoderLoader.v1 = curated_transformers.tokenization:build_char_encoder_loader_v1
-    curated-transformers.HFEncoderLoader.v1 = curated_transformers.models:build_hf_encoder_loader_v1
+    curated-transformers.HFTransformerEncoderLoader.v1 = curated_transformers.models:build_hf_encoder_loader_v1
```
danieldk, line 54: We should then probably also rename the function to `build_hf_transformer_encoder_loader_v1`.

shadeMe: Done.

danieldk (approved):

PR #120, opened 2023-03-08
Consolidate error messages and warnings into a new module

shadeMe:
## Description

This PR expands and moves all error and warning messages to a separate module like in spaCy core. A follow-up PR will add additional validation checks and their corresponding errors.

### Types of change

Refactor

## Checklist

### DIFF
--- a/curated_transformers/models/pytorch/curated_transformer.py
+++ b/curated_transformers/models/pytorch/curated_transformer.py
@@ -25,10 +25,6 @@ class CuratedTransformer(Generic[CuratedEncoderT], Module):
         super().__init__()
         self.curated_encoder = encoder
 
-        assert isinstance(
-            encoder, (AlbertEncoder, BertEncoder, RobertaEncoder)
-        ), f"Unexpected curated encoder type '{type(encoder)}'"
-
     def forward(
         self,
         input_ids: Tensor,
--- a/curated_transformers/models/pytorch/curated_transformer.py
+++ b/curated_transformers/models/pytorch/curated_transformer.py
@@ -25,6 +25,10 @@ class CuratedTransformer(Generic[CuratedEncoderT], Module):
         super().__init__()
         self.curated_encoder = encoder
 
+        assert isinstance(
+            encoder, (AlbertEncoder, BertEncoder, RobertaEncoder)
+        ), f"Unexpected curated encoder type '{type(encoder)}'"
+
     def forward(
         self,
         input_ids: Tensor,
--- a/curated_transformers/models/pytorch/hf_util.py
+++ b/curated_transformers/models/pytorch/hf_util.py
@@ -11,6 +11,7 @@ from ..._compat import transformers
 from ...errors import Errors
 
 SUPPORTED_MODEL_TYPES = ["albert", "bert", "camembert", "roberta", "xlm-roberta"]
+SUPPORTED_CURATED_ENCODERS = (AlbertEncoder, BertEncoder, RobertaEncoder)
 
 
 def _check_supported_hf_models(model_type: str):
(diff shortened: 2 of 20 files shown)


curated_transformers/models/pytorch/curated_transformer.py
```diff
@@ -25,6 +25,10 @@ def __init__(self, encoder: CuratedEncoderT) -> None:
         super().__init__()
         self.curated_encoder = encoder
 
+        assert isinstance(
+            encoder, (AlbertEncoder, BertEncoder, RobertaEncoder)
+        ), f"Unexpected curated encoder type '{type(encoder)}'"
+
```
danieldk, line 31: I am not sure if we should go here? The type already states which encoders are allowed. If we really want to do this, we should get the list of classes from the typevar constraint, so that we don't have to maintain it in two places.

curated_transformers/models/pytorch/hf_util.py
```diff
@@ -37,7 +42,11 @@ def convert_pretrained_model_for_encoder(
     elif isinstance(encoder, RobertaEncoder):
         converted = _convert_roberta_base_state(params)
     else:
-        raise ValueError(f"Unsupported encoder type: {type(encoder)}")
+        # We don't need to handle the exceptional case
+        # here as the relevant checks are already performed
+        # during the construction of the CuratedTransformer
+        # instance.
+        pass
```
danieldk, line 49: I think we should have the exception here. Would be bad to silently do nothing if the user triggered this code path somehow.

curated_transformers/tokenization/hf_loader.py
```diff
@@ -51,10 +52,20 @@ def _convert_encoder(
...
-    raise ValueError(
-        f"Loading from the '{type(tokenizer)}' huggingface tokenizer is not supported"
-    )
+    else:
+        supported_tokenizers = (
+            transformers.BertTokenizerFast,
+            transformers.RobertaTokenizerFast,
+            transformers.XLMRobertaTokenizerFast,
+            transformers.CamembertTokenizerFast,
+            transformers.BertJapaneseTokenizer,
```
danieldk, line 61: Would be nice to have this as a global 'constant' near the top of the file. Makes it easier to maintain/inspect.

curated_transformers/models/pytorch/curated_transformer.py
shadeMe: Ah, I wasn't aware that we could query types from the `TypeVar` instance.

The original thought was to catch the usage of an unsupported encoder as early as possible as a pre-condition (since the initialization of `CuratedTransformer` happens before the initialization callbacks) and have a single source of truth, i.e., this module.

We can do away with this check if you don't think it's necessary. On second thought, we don't consistently perform runtime type checking in this way elsewhere after all.

curated_transformers/models/pytorch/hf_util.py
shadeMe: Fair. I guess I'm just peeved a bit that we essentialy have to maintain the supported list of encoders in two different modules, but I don't see any way around it 🤷

curated_transformers/tokenization/hf_loader.py
shadeMe: Done.

danieldk:
@shadeMe moved it back to your board, since there is a build error.

curated_transformers/models/pytorch/curated_transformer.py
danieldk: > On second thought, we don't consistently perform runtime type checking in this way elsewhere after all.

Indeed, so I think we should just remove it for now. The type system reject these things and worst case it triggers the later exception.

shadeMe:
I've initialized the supported tokenizers global conditionally but am not sure if there's a better way to do it.

danieldk (approved):

PR #122, opened 2023-03-16
Add validation checks for piece encoders during transformer init

shadeMe:
## Description

Add additional validation checks to the transformer architecture entry points to ensure that only compatible piece encoders are used.

### Types of change

Feature

## Checklist


danieldk:
Could you explain why we want this? I think it's good to have the flexibility (in general) to use models with arbitrary encoders? E.g. I already see one combination that I currently use that would be rejected by this PR (BERT with character piece encoder, which is needed for the Japanese model).

shadeMe:
Closing this PR - Turns out I misunderstood the intention behind the [original PR](https://github.com/explosion/curated-transformers/pull/113) that introduced the configurable piece encoders.

PR #124, opened 2023-03-21
Add a workaround for invalid outputs of `nn.Linear` on MPS

danieldk:
## Description

`nn.Linear` produces incorrect outputs with certain matrix sizes when using the MPS backend:

https://github.com/pytorch/pytorch/issues/97239

The actual issue is in the underlying `torch.nn.functional.linear` function. Work around this by using an explicit matrix multiplication when the MPS backend is used.

Tested on all of the curated transformers models that I trained and after this change there are no regressions.

### Types of change

Workaround upstream issue.

## Checklist


svlandeg (approved):

danieldk:
The issue is fixed in macOS 13.3, but the workaround is even a bit faster.

PR #127, opened 2023-03-28
`tests`: Correctly handle GPU-resident tensors

shadeMe:
## Description

Tests were using `numpy` specific code to compare tensors, and this resulted in `TypeErrors` when they were executed on the GPU. This PR changes the relevant code to be device-agnostic whenever possible. In instances where this is not possible, GPU tensors are explicitly copied to the host before comparison.

Additionally, the `test_xlmr_transformer_pipe_against_hf` test was modified to use a higher tolerance for tensor comparisons to work around small differences stemming from GPU-related non-determinism.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/curated_transformers/tests/models/test_pooling.py
+++ b/curated_transformers/tests/models/test_pooling.py
@@ -1,4 +1,5 @@
-from thinc.api import Ragged, reduce_sum
+from thinc.api import Ragged, reduce_sum, NumpyOps
+from thinc.util import convert_recursive, is_cupy_array
 
 from curated_transformers.models.pooling import (
     with_ragged_last_layer,
@@ -18,14 +19,15 @@ def test_with_ragged_last_layer():
 
     Y, backprop = pooler(docs, is_train=True)
 
-    ops.xp.testing.assert_equal(
+    for y, y_h in zip(
         Y,
         [
             ops.asarray2f([[1.0], [5.0]]),
             ops.asarray2f([[9.0], [6.0]]),
             ops.asarray2f([]).reshape(0, 1),
         ],
-    )
+    ):
+        ops.xp.testing.assert_array_equal(y, y_h)
 
     dX = backprop(
         [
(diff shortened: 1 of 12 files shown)


shadeMe:
Setting to draft mode until the repo has been wired up to automatically execute GPU tests for all PRs.

honnibal:
Just testing whether marking ready triggers CI

curated_transformers/tests/models/test_hf_model.py
```diff
@@ -76,12 +76,18 @@ def test_model_against_hf_transformers(model_config):
     )
     Y_hf_encoder = hf_encoder(X, attention_mask=attention_mask)
 
-    assert torch.allclose(
-        Y_encoder.last_hidden_layer_states, Y_hf_encoder.last_hidden_state, atol=1e-6
+    torch.testing.assert_close(
+        Y_encoder.last_hidden_layer_states,
+        Y_hf_encoder.last_hidden_state,
+        atol=1e-05,
+        rtol=1e-05,
```
danieldk, line 83: Won't the default tolerances work? See the table at:

https://pytorch.org/docs/stable/testing.html#torch.testing.assert_close

I am kinda surprised that the default `rtol` would be too strict. Could this be an issue because one tensor is a CuPy tensor and it will switch to the tolerances `0.0`/`0.0` due to a `dtype` mismatch? If so, it'd probably be better to convert our tensor to Torch (or the Huggingface tensor to `ops.xp`) and use the defaults.

danieldk: Hmmm, that's a shame. Maybe we should make constants for the default tolerances (or a wrapper function), since most are the same that's probably a little cleaner than having `atol`/`rtol` everywhere. If one case deviates, we could always state the tolerances explicitly there.

danieldk (approved):

PR #137, opened 2023-05-07
Add Rotary Embeddings layer (Su et al., 2021) - take 2

danieldk:
## Description

Add the rotary embeddings layer that is used by various GPT-based LLMs.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/models/embeddings.py
+++ b/curated_transformers/models/embeddings.py
@@ -29,3 +29,84 @@ class SinusoidalPositionalEmbedding(Module):
             x - (batch, seq_len)
         """
         return self.pe[x.size(1), :]
+
+
+class RotaryEmbeddings(Module):
+    cos: Tensor
+    sin: Tensor
+    theta: Tensor
+
+    def __init__(self, width: int, *, seq_len: int = 512, base: int = 10000):
+        """Rotary embeddings (Su et al., 2021) layer. The rotary embedding
+        will be precomputed for up to 'seq _len' positions. The embedding
+        will be recomputed when a longer sequence is found in the input.
+
+        width (int):
+            Rotary embedding dimensionality, must be even.
+        seq_len (int):
+            Number of positons to initially precompute.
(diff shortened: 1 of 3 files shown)


curated_transformers/models/embeddings.py
```diff
@@ -29,3 +29,84 @@ def forward(self, x: Tensor) -> Tensor:
...
+        """
+        seq_len = input.size(-2)
+        width = self.cos.size(-1)
+        if self.cos.size(-2) < seq_len:
+            self._create_rotary_embed(width=width, length=seq_len)
+        shape = ((1,) * (input.ndim - self.cos.ndim)) + (seq_len, width)
+        rot_cos = self.cos[:seq_len, :].view(shape)
+        rot_sin = self.sin[:seq_len, :].view(shape)
+
+        # Eq 34 with ordering changed for compatibility.
```
shadeMe, line 111: Do we know the reason behind the change in ordering?

danieldk: I think this precedes the `transformers` implementation, it is also done this way in the original GPT-NeoX codebase:

https://github.com/EleutherAI/gpt-neox/blob/b608043be541602170bfcfb8ec9bf85e8a0799e0/megatron/model/positional_embeddings.py#L38

But I don't think it's really motivated. But the reasoning is probably something like:

1. If we start from Eq 34, for selecting `x_i`, `x_{i+1}`, it does not really matter what the ordering of `i` is (we could reorder the dimensions of all the parameters in the model to get either order).
2. Rotating half the array is probably more efficient than interleaved access (either when copying or when iterating with a stride).

danieldk: Found the original change through more spelunking:

https://github.com/EleutherAI/gpt-neox/pull/241

At any rate, I found out by first implementing the paper and then figuring out that it doesn't work the same way 🙈 .

PR #138, opened 2023-05-08
`Add ByteBPETokenizer`

danieldk:
## Description

This class wraps the ByteBPEProcessor class from Cutlery and does the pre/post-processing to provide output that is fit for transformer models. The class also provides two class methods to load a vocabulary, `from_files` and `from_hf_hub`
    
This change also adds an initial `Tokenizer` base class that all future tokenizers can implement.

I hope that all the abstractions hold up, but we can probably only know once we have more tokenizers.

### Types of change

Enhancement.

## Checklist


curated_transformers/tokenization/bbpe_tokenizer.py
```diff
@@ -0,0 +1,131 @@
+from typing import List, Type, TypeVar
+from cutlery import ByteBPEProcessor
+import json
+from pathlib import Path
+
+from . import PiecesWithIds, Tokenizer
```
shadeMe, line 6:
```suggestion
from .tokenizer import PiecesWithIds, Tokenizer
```
Potential circular import?

curated_transformers/tokenization/bbpe_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+
+        eos_id = self.processor.piece_id(self.eos_piece)
+        if eos_id is None:
+            raise ValueError(
+                f"Byte BPE piece encoder vocabulary doesn't contain '{self.eos_piece}' piece"
+            )
+
+        pad_id = self.processor.piece_id(self.pad_piece)
+        if eos_id is None:
```
shadeMe, line 84:
```suggestion
        if pad_id is None:
```

curated_transformers/tokenization/bbpe_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError(
+                f"Byte BPE piece encoder vocabulary doesn't contain '{self.eos_piece}' piece"
```
shadeMe, line 86:
```suggestion
                f"Byte BPE piece encoder vocabulary doesn't contain '{self.pad_piece}' piece"
```

curated_transformers/tokenization/bbpe_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+            text_lens.append(1)
+
+            ids.append(text_ids)
+            pieces.append(text_pieces)
+            lens.append(text_lens)
+
+        return PiecesWithIds(pad_id=pad_id, ids=ids, lens=lens, pieces=pieces)
+
+    @classmethod
+    def _convert_hf_tokenizer(cls: Type[Self], tokenizer) -> Self:
```
shadeMe, line 117:
```suggestion
    def _convert_hf_tokenizer(cls: Type[Self], tokenizer: Any) -> Self:
```
Or we go for the full annotation by importing `tokenizers`.

curated_transformers/tokenization/tokenizer.py
```diff
@@ -0,0 +1,72 @@
...
+        for idx, seq_ids in enumerate(self.ids):
+            mask[idx, : len(seq_ids)] = True
+        return mask
+
+    @property
+    def padded_tensor(self):
+        """Padded CPU tensor of the piece identifiers."""
+        n_seqs = len(self.ids)
+        max_len = max(len(seq_ids) for seq_ids in self.ids)
+        padded = torch.full((n_seqs, max_len), self.pad_id)
```
shadeMe, line 36: I think this defaults to `torch.int64`. Given that 32-bits is plenty for the IDs, we might want to explicitly set `dtype=torch.int32` for better throughput on the GPU (when it eventually gets uploaded).

danieldk: Makes sense, I'll fix it (`int32` didn't work in the past for a lot of things like the embedding layer in the past, to I built up this habit).

danieldk: Fixed.

shadeMe (approved):

PR #139, opened 2023-05-08
Factor self attention out of the bert package

danieldk:
## Description

The same mechanism is reusable between transformer models. While at it, add some docstrings and make the attention mask optional.

### Types of change

Refactor

## Checklist

### DIFF
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -4,8 +4,6 @@ import torch
 from torch import Tensor
 from torch.nn import Linear, Module
 
-from ..errors import Errors
-
 
 class AttentionMask:
     bool_mask: Tensor
@@ -107,9 +105,8 @@ class SelfAttention(Module):
         self.num_heads = num_attention_heads
         if self.model_dim % self.num_heads != 0:
             raise ValueError(
-                Errors.E003.format(
-                    hidden_width=self.model_dim, num_heads=self.num_heads
-                )
+                f"The hidden width of the transformer ({self.model_dim}) must be "
+                "divisible by the number of self-attention heads ({self.num_heads})"
             )
 
         self.dims_per_head = self.model_dim // self.num_heads
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -149,19 +149,6 @@ class SelfAttention(Module):
         return out
 
 
-def combine_heads(x: Tensor) -> Tensor:
-    """
-    Combine the attention head representations. The inverse of
-    'split_heads'.
-
-    Shapes:
-        x - (batch, head, seq_len, width_per_head)
-        output - (batch, seq_len, width)
-    """
-    batch_size, head, seq_len, model_dim = x.size()
-    return x.transpose(1, 2).contiguous().view(batch_size, seq_len, head * model_dim)
-
-
 def split_heads(x: Tensor, num_heads: int) -> Tensor:
     """
     Split the input by attention head. The caller must validate
@@ -182,3 +169,16 @@ def split_heads(x: Tensor, num_heads: int) -> Tensor:
     dims_per_head = model_dim // num_heads
 
     return x.view(batch_size, seq_len, num_heads, dims_per_head).transpose(1, 2)
+
+
+def combine_heads(x: Tensor) -> Tensor:
+    """
+    Combine the attention head representations. The inverse of
+    'split_heads'.
+
+    Shapes:
+        x - (batch, head, seq_len, width_per_head)
+        output - (batch, seq_len, width)
+    """
+    batch_size, head, seq_len, model_dim = x.size()
+    return x.transpose(1, 2).contiguous().view(batch_size, seq_len, head * model_dim)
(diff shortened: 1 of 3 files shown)


curated_transformers/models/attention.py
```diff
@@ -44,15 +44,24 @@ def __init__(self, *, dropout_prob: float = 0.1):
...
         """
+        Apply attention layer to the given key, query and value.
+
+        Sequence elements that are marked with `False` in the attention mask
+        are ignored by the attention mechanism (if a mask is provided).
+
+        k (Tensor): key.
+        q (Tensor): query.
+        v (Tensor): value.
+        attn_mask (Optional[AttentionMask]): attention mask.
```
shadeMe, line 58:
```suggestion
        attn_mask (Optional[AttentionMask]): attention mask.
        
```

curated_transformers/models/attention.py
```diff
@@ -61,12 +70,112 @@ def forward(
...
+        dropout_prob: float = 0.1,
+        hidden_width: int = 768,
+        num_attention_heads: int = 12,
+    ):
+        """Construct a self-attention layer.
+
+        dropout_prob (float): dropout to apply between the self-attention
+            and output layers (default: 0.1).
+        hidden_width (int): hidden width of the layer (default: 768).
+        num_attention_heads (int): number of attenion heads (default: 12).
```
shadeMe, line 101:
```suggestion
        num_attention_heads (int): number of attention heads (default: 12).
```

curated_transformers/models/attention.py
```diff
@@ -61,12 +70,112 @@ def forward(
...
 
         attn_weights = attn_scores.softmax(dim=-1)
         attn_values = self.dropout(attn_weights @ v)
 
         return attn_values
+
+
+# https://www.tensorflow.org/text/tutorials/transformer#multi-head_attention
+class SelfAttention(Module):
+    """Transformer self-attention layer (Vaswani et al., 2016)."""
```
shadeMe, line 87:
```suggestion
    """Transformer self-attention layer (Vaswani et al., 2017)."""
```

curated_transformers/models/attention.py
```diff
@@ -61,12 +70,112 @@ def forward(
...
+        super().__init__()
+
+        self.model_dim = hidden_width
+        self.num_heads = num_attention_heads
+        if self.model_dim % self.num_heads != 0:
+            raise ValueError(
+                Errors.E003.format(
+                    hidden_width=self.model_dim, num_heads=self.num_heads
+                )
+            )
```
shadeMe, line 112: Shall we inline this error here already?

curated_transformers/models/attention.py
```diff
@@ -44,15 +44,24 @@ def __init__(self, *, dropout_prob: float = 0.1):
...
+        Sequence elements that are marked with `False` in the attention mask
+        are ignored by the attention mechanism (if a mask is provided).
+
+        k (Tensor): key.
+        q (Tensor): query.
+        v (Tensor): value.
+        attn_mask (Optional[AttentionMask]): attention mask.
         Shapes:
             k, q, v - (batch, heads, seq_len, width)
             attn_mask - (batch, seq_len)
```
shadeMe, line 61: I think it makes sense to also document the output shape (in all modules)?

curated_transformers/models/attention.py
```diff
@@ -61,12 +70,112 @@ def forward(
...
+        x (Tensor): input to apply self-attention to.
+        attn_mask (Optional[AttentionMask]): attention mask.
+
+        Shapes:
+            x - (batch, seq_len, width)
+            attn_mask - (batch, seq_len)
+        """
+
+        proj = self.input(x)
+        q, k, v = proj.chunk(3, dim=-1)
```
shadeMe, line 135: Perhaps a comment here to explain why we're doing the projection just once.

curated_transformers/models/attention.py
```diff
@@ -61,12 +70,112 @@ def forward(
...
+    """
+    Combine the attention head representations. The inverse of
+    'split_heads'.
+
+    Shapes:
+        x - (batch, head, seq_len, width_per_head)
+        output - (batch, seq_len, width)
+    """
+    batch_size, head, seq_len, model_dim = x.size()
+    return x.transpose(1, 2).contiguous().view(batch_size, seq_len, head * model_dim)
```
shadeMe, line 159: This reads better if it came after `split_heads`, given that it's the inverse of the former.

curated_transformers/models/attention.py
danieldk: Inlined.

curated_transformers/models/attention.py
danieldk: Done for the attention module.

curated_transformers/models/attention.py
danieldk: Done.

curated_transformers/models/attention.py
danieldk: Indeed, fixed!

PR #153, opened 2023-05-12
Improved Hugging Face Hub model loading infra, apply to RoBERTa

danieldk:
## Description

This change is the first in a series of changes that moves HF model loading from an idempotent function to the models themselves. The idempotent function got quite convoluted when adding decoder architectures, furthermore it probably makes more sense to couple model-specifics to the model. This change introduces the `FromPretrainedHFModel` mixin. Modules that want support for loading parameters from Hugging Face Hub can derive from this mixin and implement the `convert_hf_state_dict` and `from_hf_config` methods. The mixin will then provide the `from_hf_hub` method to download a model from the Hugging Face Hub.

This change also migrates the RoBERTa models to use the mixin.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/models/hf_hub.py
+++ b/curated_transformers/models/hf_hub.py
@@ -76,7 +76,7 @@ class FromPretrainedHFModel(ABC):
     def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True):
         """Load a state dictionary.
 
-        This method is automatically impemented by also deriving from
+        This method is automatically implemented by also deriving from
         `torch.nn.Module`. This mixin does not derive from `Module` in
         order to be an abstract base class.
         """
--- a/curated_transformers/models/hf_hub.py
+++ b/curated_transformers/models/hf_hub.py
@@ -29,8 +29,8 @@ class FromPretrainedHFModel(ABC):
         """Convert a state dict of a Hugging Face model to a valid
         state dict for the module.
 
-        params (Mapping[str, Parameter]): the state dict to convert.
-        RETURNS (Mapping[str, Tensor]): the converted state dict.
+        params (Mapping[str, Parameter]): The state dict to convert.
+        RETURNS (Mapping[str, Tensor]): The converted state dict.
         """
         raise NotImplementedError
 
@@ -41,7 +41,7 @@ class FromPretrainedHFModel(ABC):
         model configuration.
 
         hf_config (Any): Hugging Face model configuration.
-        RETURNS (Self): module constructed using the configuration.
+        RETURNS (Self): Module constructed using the configuration.
         """
         raise NotImplementedError
 
--- a/curated_transformers/models/hf_hub.py
+++ b/curated_transformers/models/hf_hub.py
@@ -1,0 +1,74 @@
+from typing import Any, Mapping, Type, TypeVar
+from abc import ABC, abstractmethod
+from torch import Tensor
+from torch.nn import Parameter
+
+from .._compat import has_hf_transformers, transformers
+from .module import CausalLMModule, DecoderModule, EncoderModule
+
+
+# Only provided as typing.Self in Python 3.11+.
+Self = TypeVar("Self", bound="FromPretrainedHFModel")
+
+
+class FromPretrainedHFModel(ABC):
+    """Mixin class for downloading models from Hugging Face Hub.
+
+    A module using this mixin can implement the `convert_hf_state_dict`
+    and `from_hf_config` methods. The mixin will then provide the
+    `from_hf_hub` method to download a model from the Hugging Face Hub.
+    """
+
+    @classmethod
+    @abstractmethod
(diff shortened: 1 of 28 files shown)


danieldk:
I'll put this in draft for a moment, I think it should be fairly easy to do this with the `huggingface_hub` package, without `transformers`.

danieldk:
Done, no more `transformers` needed to load a model. We should do the same for the tokenizers.

curated_transformers/models/hf_hub.py
```diff
@@ -0,0 +1,83 @@
...
+            repo_id=name, filename=HF_MODEL_CONFIG, revision=revision
+        )
+        with open(config_filename, "r") as f:
+            config = json.load(f)
+        model = cls.from_hf_config(hf_config=config)
+
+        # Download model and convert HF parameter names to ours.
+        model_filename = hf_hub_download(
+            repo_id=name, filename=HF_MODEL_CHECKPOINT, revision=revision
+        )
```
shadeMe, line 67: Does this work with sharded checkpoints like the ones for GPT NeoX? Those split the parameters across different files, each of which still works as a single PyTorch checkpoint. There's [extra code](https://github.com/huggingface/transformers/blob/9a50cb6195cb477c1feae8f0fb2bce20b679246f/src/transformers/modeling_utils.py#L351) in `transformers` to handle this for LLMs.

curated_transformers/models/roberta/_hf.py
```diff
@@ -0,0 +1,79 @@
...
+from torch import Tensor
+from torch.nn import Parameter
+
+from .config import RobertaConfig
+from ..hf_util import _merge_qkv
+
+
+def convert_hf_config(hf_config: Any) -> RobertaConfig:
+    # WARNING: seem like we cannot get the max length from the model? Maybe
+    # we should do the max position embeddings minus 2?
```
shadeMe, line 12: Yeah, I think that's part of the tokenizer for some odd reason. We should be able to deserialize its config along with the model's, I think?

curated_transformers/models/hf_hub.py
```diff
@@ -0,0 +1,83 @@
...
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def from_hf_config(cls: Type[Self], *, hf_config: Any) -> Self:
+        """Create the module from a Hugging Face model JSON-deserialized
+        model configuration.
+
+        hf_config (Any): Hugging Face model configuration.
+        RETURNS (Self): module constructed using the configuration.
```
shadeMe, line 44:
```suggestion
        RETURNS (Self): Module constructed using the configuration.
```

curated_transformers/models/hf_hub.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    @abstractmethod
+    def convert_hf_state_dict(
+        cls, params: Mapping[str, Parameter]
+    ) -> Mapping[str, Tensor]:
+        """Convert a state dict of a Hugging Face model to a valid
+        state dict for the module.
+
+        params (Mapping[str, Parameter]): the state dict to convert.
```
shadeMe, line 32:
```suggestion
        params (Mapping[str, Parameter]): The state dict to convert.
```

curated_transformers/models/hf_hub.py
```diff
@@ -0,0 +1,83 @@
...
+        RETURNS (Mapping[str, Tensor]): the converted state dict.
```
shadeMe, line 33:
```suggestion
        RETURNS (Mapping[str, Tensor]): The converted state dict.
```

curated_transformers/models/hf_hub.py
```diff
@@ -0,0 +1,83 @@
...
+        return model
+
+    @abstractmethod
+    def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True):
+        """Load a state dictionary.
+
+        This method is automatically impemented by also deriving from
+        `torch.nn.Module`. This mixin does not derive from `Module` in
+        order to be an abstract base class.
+        """
```
shadeMe, line 82: Missing docstrings for the parameters.

curated_transformers/models/hf_hub.py
danieldk: No, it doesn't. We should put sharded checkpoints on the todo list.

curated_transformers/models/roberta/_hf.py
shadeMe: We can still pass the tokenizer config as an `Optional`, use its `model_max_length` attribute when available and falling back to the position embeddings when it's missing? That seems a bit more robust than always using the latter.

shadeMe: After internal discussion, we'll keep the current implementation as the model should be seen as the source of truth for `max_length`, and to reduce coupling between models and their tokenizers.

PR #154, opened 2023-05-14
Add support for decoding in Byte BPE tokenizers

danieldk:
## Description

This adds the `decode` method to tokenizers and implements it for the Byte BPE tokenizers (RoBERTa and GPT-NeoX). For consistency, the `__call__` method of tokenizer is renamed to `encode` and `__call__` wraps `encode`.

### Types of change

Feature.

## Checklist


curated_transformers/tokenization/roberta_tokenizer.py
```diff
@@ -1,25 +1,59 @@
...
+        if bos_id is None:
+            raise ValueError(
+                f"RoBERTa piece encoder vocabulary doesn't contain '{self.bos_piece}' piece"
+            )
+
+        eos_id = self.processor.piece_id(self.eos_piece)
+        if eos_id is None:
+            raise ValueError(
+                f"RoBERTa piece encoder vocabulary doesn't contain '{self.eos_piece}' piece"
+            )
```
shadeMe, line 43: Could we move these checks to the ctor? Or is it valid to pass an uninitialized/empty processor to it?

On a different note, we could move these checks into their own function and call them both in the pre-/post-decoder.

danieldk: Yeah, I think we need to allow to create blank tokenizers for `spacy-curated-transformers`. Though maybe we should make things nice here and hack around that later in `spacy-curated-transformers`? Even more with spaCy ❤️ PyTorch?

I'll factor these out to separate functions.

shadeMe: I don't have a strong opinion as such, but having these checks happen inside the ctor in `curated-transformers` would indeed be nicer.

danieldk: Fixed.

shadeMe (approved):

PR #157, opened 2023-05-15
Add support for sharded HF checkpoints

shadeMe:
## Description

This PR adds support for down(loading) models from HF Hub that are stored as sharded checkpoints (currently just LLM models, I think). Tested locally with `decapoda-research/llama-7b-hf`.

The default implementation simply combines the state dicts of individual shards into a single instance and passes that to the model. This will change in upcoming PRs once we start implementing logic to load pre-trained weights in a memory-conservative manner.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/models/hf_hub.py
+++ b/curated_transformers/models/hf_hub.py
@@ -1,13 +1,16 @@
 import json
-from typing import Any, Mapping, Type, TypeVar
+from typing import Any, Iterable, Mapping, Type, TypeVar
 from abc import ABC, abstractmethod
 from huggingface_hub import hf_hub_download
+from requests import HTTPError
 import torch
 from torch import Tensor
 from torch.nn import Parameter
 
 HF_MODEL_CONFIG = "config.json"
 HF_MODEL_CHECKPOINT = "pytorch_model.bin"
+HF_MODEL_SHARDED_CHECKPOINT_INDEX = "pytorch_model.bin.index.json"
+HF_MODEL_SHARDED_CHECKPOINT_INDEX_WEIGHTS_KEY = "weight_map"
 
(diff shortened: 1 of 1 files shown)


curated_transformers/models/hf_hub.py
```diff
@@ -86,3 +80,71 @@ def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True):
...
+        )
+    except:
+        raise ValueError(
+            f"Couldn't find a valid config for model `{name}` "
+            f"(revision `{revision}`) on HuggingFace Model Hub"
+        )
+
+
+def _get_model_checkpoint_filepaths(name: str, revision: str) -> Iterable[str]:
+    # Attempt to download a non-sharded checkpoint first.
```
danieldk, line 98: I was wondering if we don't want to try getting a sharded model first. If haven't looked into this yet, but I think that one of the benefits of shard loading is that in the future we could e.g. quantize the parameters by shard when loading large models. As a result we never need to have the full FP32 model in memory.

danieldk (commented):
It really great to have this, but we definitely need a test. From a quick look it seems like the HF random test models currently only have sharded codegen models (that we don't support yet). Maybe we should make this testing org that we talked about on HF hub and up a sharded RoBERTa there?

curated_transformers/models/hf_hub.py
shadeMe: Yeah, that was my thinking too. That being said, the order in which we check/get the filepaths shouldn't matter as sharded and non-sharded checkpoints are mutually exclusive (as far as I can tell) - The `python_model.bin` file is only present in the case of the latter.

danieldk (approved):

PR #158, opened 2023-05-16
Improve the API sections, add more classes

danieldk:
## Description

Improve the API sections, add more classes.  Also infer parameter type documentation from type annotations.

### Types of change

Documentation improvements.

## Checklist

### DIFF
--- a/curated_transformers/models/embeddings.py
+++ b/curated_transformers/models/embeddings.py
@@ -42,11 +42,11 @@ class RotaryEmbeddings(Module):
         will be precomputed for up to 'seq _len' positions. The embedding
         will be recomputed when a longer sequence is found in the input.
 
-        width (int):
+        :param width:
             Rotary embedding dimensionality, must be even.
-        seq_len (int):
+        :param seq_len:
             Number of positons to initially precompute.
-        base (int):
+        :param base:
             The base used for Θ_i, determines the cycle length of the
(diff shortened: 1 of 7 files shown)


curated_transformers/models/embeddings.py
```diff
@@ -94,10 +94,10 @@ def forward(self, input: torch.Tensor, *, positions: Optional[Tensor] = None):
         """
         Apply rotary embeddings to an array.
 
-        input (Tensor): Array to apply the rotary embeddings to.
-        positions (Optional[Tensor]): positions of the inputs. If no
-            positions are provided, they are assumed to be [0, seq_len).
-        RETURNS (Tensor): Array with the rotary embeddings applied.
+        :param input: Array to apply the rotary embeddings to.
```
danieldk, line 97: Had to fix this up already, because I was getting errors when generating the documentation.

shadeMe: This means that the docs generator will automatically pull the type and default values from the function signature?

danieldk: Not sure what is happening with the constructor documentation, let me check...

danieldk: Better:

Screenshot 2023-05-16 at 11 13 16

shadeMe (approved):

PR #166, opened 2023-05-19
`BertTokenizer` and friends

shadeMe:
## Description

Add `BertTokenizer` and companion classes. 

### Types of change

New feature

## Checklist


curated_transformers/tests/tokenization/test_bert_tokenizer.py
```diff
@@ -0,0 +1,243 @@
...
+@pytest.fixture
+def toy_tokenizer(test_dir):
+    return BertTokenizer.from_files(
+        vocab_path=test_dir / "toy.wordpieces",
+    )
+
+
+@pytest.mark.skipif(not has_hf_transformers, reason="requires huggingface transformers")
+@pytest.mark.slow
+def test_berttokenizer_hf_tokenizer(sample_texts):
```
danieldk, line 21: We now have this test in three places, maybe we should factor it out to a function that can be called with the model name? Then if we want to check other things later like padding or attention masks, we could easily add them to all the places.

curated_transformers/tests/tokenization/test_bert_tokenizer.py
```diff
@@ -0,0 +1,243 @@
...
+    hf_pieces = hf_tokenizer(sample_texts)
+
+    assert pieces.ids == hf_pieces["input_ids"]
+
+
+def test_bert_toy_tokenizer(toy_tokenizer, short_sample_texts):
+    encoding = toy_tokenizer(short_sample_texts)
+    _check_toy_tokenizer(encoding)
+
+    assert toy_tokenizer.decode(encoding.ids) == [
```
danieldk, line 35: It would be nice to have a roundtrip encode/decode test as well (similar to the GPT-NeoX and RoBERTa tokenizer tests), if possible.

shadeMe: As discussed internally, I've added round-trip checks to the HF tokenizer comparison.

danieldk (approved):

PR #172, opened 2023-05-24
Add support for left padding

danieldk:
## Description

Left-padding is not strictly needed for transformer models, since the attention mask determines what tokens are attended to. However, left-padded sequences are slightly easier to reason about/debug in generation, since the input tokens immediately precede the generated tokens.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/tokenization/tokenizer.py
+++ b/curated_transformers/tokenization/tokenizer.py
@@ -2,6 +2,7 @@ from typing import Iterable, List, Optional, TypeVar, Union
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 import torch
+from torch import Tensor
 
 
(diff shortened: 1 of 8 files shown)


shadeMe (approved):

curated_transformers/tokenization/tokenizer.py
```diff
@@ -13,25 +14,46 @@ class PiecesWithIds:
     ids: List[List[int]]
     pieces: List[List[str]]
 
-    @property
-    def attention_mask(self):
-        """CPU tensor with attention masks.
+    def attention_mask(self, pad_left: bool = False) -> Tensor:
```
shadeMe, line 17:
```suggestion
    def attention_mask(self, *, pad_left: bool = False) -> Tensor:
```

PR #173, opened 2023-05-24
Allow more fine-grained control over tokenizer input

danieldk:
## Description

Before this change, tokenizers took `List[str]` is their input. However, this is not ideal in some cases. For instance in many LLMs we have to provide a model with special pieces to signal e.g. the start/end of a prompt. To serve these use cases, this change allows more fine-grained control over the input.

Besides supporting `List[str]`, the tokenizers now also support `List[InputChunks]`. `InputChunks` is a list-derived data structure that can contain one of two types of chunks:

* `TextChunk`: this is a regular piece of text that should be tokenized.
* `SpecialPieceChunk`: this type of chunk stores a single piece that should be looked up in directly in the vocabulary. Additionally, it can have strings to be appended/prepended to the preceding/succeeding text chunks.

The tokenizer normalizes these chunks by merging consecutive text chunks and adding the strings specified in special piece chunks. After this merge step, it processes the chunks.

Since we can have special pieces in the input with this change, begin/end of sequence markers can now be added safely in the pre-encoders.

In order to test the use of special piece chunks in the input, the GPT-NeoX tokenizer is updated to include special pieces in the vocabulary. We may eventually want to do this for other types of tokenizers as well.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/tokenization/bbpe_tokenizer.py
+++ b/curated_transformers/tokenization/bbpe_tokenizer.py
@@ -1,6 +1,11 @@
 from typing import Iterable, List
 from curated_tokenizers import ByteBPEProcessor
 
+from curated_transformers.tokenization.chunks import (
+    MergedInputChunks,
+    MergedSpecialPieceChunk,
+)
+
 from .tokenizer import PiecesWithIds, Tokenizer
 
 
@@ -22,22 +27,30 @@ class ByteBPETokenizer(Tokenizer):
     def _decode(self, input: Iterable[Iterable[int]]) -> List[str]:
         return [self.processor.decode_from_ids(ids) for ids in input]
 
-    def _encode(self, input: Iterable[str]) -> PiecesWithIds:
+    def _encode(self, input: Iterable[MergedInputChunks]) -> PiecesWithIds:
         ids = []
         pieces = []
 
-        for text in input:
-            text_ids = []
-            text_pieces = []
+        for seq in input:
+            seq_ids = []
+            seq_pieces = []
 
-            for idx, token in enumerate(text.split(" ")):
-                if idx != 0:
-                    token = " " + token
-                token_ids, token_pieces = self.processor.encode(token)
-                text_ids.extend(token_ids)
-                text_pieces.extend(token_pieces)
+            for chunk in seq:
+                if isinstance(chunk, MergedSpecialPieceChunk):
+                    piece_id = self.processor.piece_id(chunk.piece)
+                    if piece_id is None:
+                        raise ValueError(f"Unknown special piece: {chunk.piece}")
+                    seq_ids.append(piece_id)
+                    seq_pieces.append(chunk.piece)
+                else:
+                    for idx, token in enumerate(chunk.text.split(" ")):
+                        if idx != 0:
+                            token = " " + token
+                        token_ids, token_pieces = self.processor.encode(token)
+                        seq_ids.extend(token_ids)
+                        seq_pieces.extend(token_pieces)
 
-            ids.append(text_ids)
-            pieces.append(text_pieces)
+            ids.append(seq_ids)
+            pieces.append(seq_pieces)
 
         return PiecesWithIds(ids=ids, pieces=pieces)
--- a/curated_transformers/tokenization/chunks.py
+++ b/curated_transformers/tokenization/chunks.py
@@ -80,7 +80,7 @@ class InputChunks(List[ChunkT]):
                     new_chunks[-1].text += chunk.text  # type: ignore[union-attr]
                 else:
                     new_chunks.append(dataclasses.replace(chunk))
-            else:
+            elif isinstance(chunk, SpecialPieceChunk):
                 if chunk.before:
                     if last_is_text:
                         new_chunks[-1].text += chunk.before  # type: ignore[union-attr]
@@ -89,5 +89,7 @@ class InputChunks(List[ChunkT]):
                 new_chunks.append(MergedSpecialPieceChunk(chunk.piece))
                 if chunk.after:
                     new_chunks.append(TextChunk(chunk.after))
+            else:
+                raise ValueError(f"Unknown chunk type: {type(chunk).__name__}")
 
         return new_chunks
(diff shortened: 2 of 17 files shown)


danieldk:
> Looks like I am relying on features that are too now, I'll check if there is an easy fix.

`typing` saves the day.

curated_transformers/tokenization/chunks.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        new_chunks = MergedInputChunks()
+        for chunk in self:
+            last_is_text = new_chunks and isinstance(new_chunks[-1], TextChunk)
+            if isinstance(chunk, TextChunk):
+                if last_is_text:
+                    new_chunks[-1].text += chunk.text  # type: ignore[union-attr]
+                else:
+                    new_chunks.append(dataclasses.replace(chunk))
+            else:
```
shadeMe, line 83: It'd be a good idea to do a `all((isinstance(x, (SpecialPieceChunk, TextChunk) for x in self))` at the beginning to check if all the elements are of the expected type(s).

curated_transformers/tokenization/roberta_tokenizer.py
```diff
@@ -4,9 +4,10 @@
 from pathlib import Path
 
 from .bbpe_tokenizer import ByteBPETokenizer
+from .chunks import InputChunks, SpecialPieceChunk, TextChunk
```
shadeMe, line 7:
```suggestion
```

curated_transformers/tokenization/roberta_tokenizer.py
```diff
@@ -4,9 +4,10 @@
...
 from .hf_hub import FromPretrainedHFTokenizer
-from .tokenizer import PiecesWithIds, PostEncoder, PreDecoder
-from .util import remove_pieces_from_sequence, add_bos_eos_to_encoding
+from .tokenizer import AddBosEosPreEncoder, PreEncoder, PreDecoder
```
shadeMe, line 9:
```suggestion
from .tokenizer import AddBosEosPreEncoder, PreDecoder
```

curated_transformers/tokenization/gpt_neox_tokenizer.py
```diff
@@ -39,7 +39,11 @@ def convert_hf_tokenizer(cls: Type[Self], tokenizer: Any) -> Self:
         deserialized = json.loads(serialized)
         vocab_merges = deserialized["model"]
         merges = [tuple(merge.split(" ")) for merge in vocab_merges["merges"]]
-        processor = ByteBPEProcessor(vocab_merges["vocab"], merges)
+        vocab = vocab_merges["vocab"]
```
shadeMe, line 42: Don't we need to add this logic to other `ByteBPETokenizer` subclasses as well?

curated_transformers/tokenization/bbpe_tokenizer.py
```diff
@@ -22,22 +27,30 @@ def __init__(
...
-                text_pieces.extend(token_pieces)
-
-            ids.append(text_ids)
-            pieces.append(text_pieces)
+        for seq in input:
+            seq_ids = []
+            seq_pieces = []
+
+            for chunk in seq:
+                if isinstance(chunk, MergedSpecialPieceChunk):
```
shadeMe, line 39: Do we want to add validate the iterable's elements' type here (and in other tokenizers) as well?

docs/source/tokenizers.rst
```diff
@@ -1,6 +1,30 @@
 Tokenization
 ============
 
+Tokenizer inputs
+----------------
+
+Each tokenizer accepts a ``List[str]`` or a ``List[InputChunks]``. In
+most cases passing a list of strings suffices. However passing
```
shadeMe, line 8:
```suggestion
most cases, passing a list of strings should suffice. However, passing
```

docs/source/tokenizers.rst
```diff
@@ -1,6 +1,30 @@
...
+:class:`.InputChunks` can be useful when adding special pieces to the input is
+required.
```
shadeMe, line 10:
```suggestion
:class:`.InputChunks` can be useful when special pieces need to be added
to the input.
```

docs/source/tokenizers.rst
```diff
@@ -1,6 +1,30 @@
...
+
+When the tokenizer is called with a list of strings, each string is
+automatically converted to a :class:`.TextChunk`, which represents a text chunk
+that should be tokenized. The other type of supported chunk is the
+:class:`.SpecialPieceChunk`. The piece stored by this type of chunk is not
+tokenized, but looked up directly.
```
shadeMe, line 16:
```suggestion
:class:`.SpecialPieceChunk`. The piece stored by this type of chunk is not
tokenized but looked up directly.
```

curated_transformers/tokenization/chunks.py
```diff
@@ -0,0 +1,93 @@
...
+        in special piece chunks.
+        """
+        new_chunks = MergedInputChunks()
+        for chunk in self:
+            last_is_text = new_chunks and isinstance(new_chunks[-1], TextChunk)
+            if isinstance(chunk, TextChunk):
+                if last_is_text:
+                    new_chunks[-1].text += chunk.text  # type: ignore[union-attr]
+                else:
+                    new_chunks.append(dataclasses.replace(chunk))
```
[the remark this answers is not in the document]
danieldk: Yeah, that's apparently the standard way to copy with `dataclasses`, but it's quite handy, you can optionally add replacements as kwargs.

curated_transformers/tokenization/gpt_neox_tokenizer.py
danieldk: Definitely. I want to do this in a follow up PR, where we also switch to `tokenizer.json` for the non-sentencepiece tokenizers.

curated_transformers/tokenization/bbpe_tokenizer.py
shadeMe: Ah, right. That's true.

shadeMe (approved):

PR #176, opened 2023-05-24
Load pre-trained models from HF Hub without unnecessary allocations

shadeMe:
## Description

This PR prevents the default initialization of models whose parameters will be replaced with those from HuggingFace Hub immediately. Furthermore, the loading of the serialized weights is performed in a zero-copy fashion when possible.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/models/embeddings.py
+++ b/curated_transformers/models/embeddings.py
@@ -74,7 +74,7 @@ class RotaryEmbeddings(Module):
 
         # Ignore allocations on the meta device as we don't persist our buffer,
         # i.e., we don't expect the backing tensor to be replaced with pretrained weights.
-        if device.type == "meta":
+        if device is not None and device.type == "meta":
             device = None
         # Θ_i = 10000^(-2(i-1)/d)
         theta = torch.pow(
--- a/curated_transformers/models/embeddings.py
+++ b/curated_transformers/models/embeddings.py
@@ -72,6 +72,10 @@ class RotaryEmbeddings(Module):
         if width % 2:
             raise ValueError(f"Width of rotary embeddings must be even, was: {width}")
 
+        # Ignore allocations on the meta device as we don't persist our buffer,
+        # i.e., we don't expect the backing tensor to be replaced with pretrained weights.
+        if device == torch.device("meta"):
+            device = None
         # Θ_i = 10000^(-2(i-1)/d)
         theta = torch.pow(
             base, -torch.arange(0, width, 2, dtype=torch.float, device=device) / width
--- a/curated_transformers/models/embeddings.py
+++ b/curated_transformers/models/embeddings.py
@@ -7,13 +7,22 @@ from torch.nn import Module
 
 # https://pytorch.org/tutorials/beginner/transformer_tutorial.html
 class SinusoidalPositionalEmbedding(Module):
-    def __init__(self, dim: int, max_len: int, *, normalize=True):
+    def __init__(
+        self,
+        dim: int,
+        max_len: int,
+        *,
+        normalize=True,
+        device: Optional[torch.device] = None,
+    ):
         super().__init__()
 
-        position = torch.arange(max_len).unsqueeze(1)
-        div_term = torch.exp(torch.arange(0, dim, 2) * (-math.log(10000.0) / dim))
+        position = torch.arange(max_len, device=device).unsqueeze(1)
+        div_term = torch.exp(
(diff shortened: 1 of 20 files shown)


curated_transformers/models/embeddings.py
```diff
@@ -48,14 +64,22 @@ def __init__(self, width: int, *, seq_len: int = 512, base: int = 10000):
...
+        :param device: Device on which the module is to be initialized.
+        """
         super().__init__()
 
         if width % 2:
             raise ValueError(f"Width of rotary embeddings must be even, was: {width}")
 
+        # Ignore allocations on the meta device as we don't persist our buffer,
+        # i.e., we don't expect the backing tensor to be replaced with pretrained weights.
+        if device == torch.device("meta"):
```
danieldk, line 77:
```suggestion
        if device.type == "meta":
```

curated_transformers/models/loading_util.py
```diff
@@ -0,0 +1,105 @@
...
+
+    size_mismatch_msg = (
+        "Expected the size of the replacement for `{}` to be {}, but got {}"
+    )
+
+    if is_parameter:
+        old_param = module._parameters[tensor_name]
+        assert old_param is not None
+        if old_param.shape != tensor.shape:
+            raise ValueError(
```
danieldk, line 38: Do we also need to check that the dtype matches?

curated_transformers/models/loading_util.py
```diff
@@ -0,0 +1,105 @@
...
+                size_mismatch_msg.format(tensor_name, old_param.shape, tensor.shape)
+            )
+        new_param = Parameter(tensor, requires_grad=old_param.requires_grad)
+        module._parameters[tensor_name] = new_param
+    else:
+        old_buffer = module._buffers[tensor_name]
+        assert old_buffer is not None
+        if old_buffer.shape != tensor.shape:
```
danieldk, line 46: Same here (dtype).

curated_transformers/models/loading_util.py
```diff
@@ -0,0 +1,105 @@
...
+                raise ValueError(f"Key `{name}` not found in module `{prefix}`")
+            replacement = state_dict[key]
+            _emplace_module_tensor(module, name, replacement, device=device)
+
+    def traverse(module: Module, state_dict: Mapping[str, torch.Tensor], prefix: str):
+        emplace(module, state_dict, prefix)
+
+        for name, child in module._modules.items():
+            if child is not None:
+                child_prefix = prefix + name + "."
```
danieldk, line 97:
```suggestion
                child_prefix = f"{prefix}{name}."
```

curated_transformers/models/loading_util.py
```diff
@@ -0,0 +1,105 @@
...
+                child_state_dict = {
+                    k: v for k, v in state_dict.items() if k.startswith(child_prefix)
+                }
+                traverse(child, child_state_dict, child_prefix)
+
+    traverse(module, state_dict, "")
+    del traverse
+    del emplace
```
danieldk, line 105: Why are these needed? I could understand the deletion of `emplace` immediately after the `emplace` call (to avoid keeping it in memory while traversing, but doesn't do the interpreter the necessary analysis?)

danieldk (commented):
Only some small comments/questions.

curated_transformers/models/loading_util.py
shadeMe: Good question. It's probably a good idea, but once implement quantized weight loading, the types won't match anyway. So, I think we can dispense with it?

curated_transformers/models/loading_util.py
shadeMe: I was just following [what `Module.load_state_dict` does](https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.load_state_dict), to be honest. I assumed it was to make sure that the `state_dict` wouldn't stick around, but it wouldn't make a difference in our case.

curated_transformers/models/loading_util.py
shadeMe: Added the check.

curated_transformers/models/loading_util.py
shadeMe: Rewrote the function to use a queue instead (and added some extra sanity checks).

danieldk (approved):

curated_transformers/models/loading_util.py
danieldk: Much nicer!

PR #181, opened 2023-06-05
Add support for normalization in `Tokenizer`

shadeMe:
## Description

Also fix a bug where the incorrect key was being used to fetch the `strip_accents` value from the BERT tokenizer's JSON config.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/tokenization/tokenizer.py
+++ b/curated_transformers/tokenization/tokenizer.py
@@ -3,6 +3,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 import torch
 from torch import Tensor
+import unicodedata
 
 from .chunks import InputChunks, MergedInputChunks, SpecialPieceChunk, TextChunk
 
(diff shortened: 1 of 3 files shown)


curated_transformers/tokenization/tokenizer.py
```diff
@@ -205,3 +217,51 @@ def __call__(self, chunks: Iterable[InputChunks]) -> List[InputChunks]:
...
+
+class DefaultNormalizer(Normalizer):
+    """
+    Performs normalization operations on input chunks before encoding.
+    """
+
+    def __init__(
+        self,
+        *,
+        utf_normalization: Optional[str] = None,
```
danieldk, line 230: Since there are only four forms, maybe we want an enum? If not, we should maybe enumerate the options in the docs.

danieldk (commented):
One small question/comment.

danieldk (approved):

PR #183, opened 2023-06-06
Validate tokenizer input chunks

shadeMe:
## Description

Validate input chunks passed to the tokenizer to idenfity invalid special tokens and prevent special tokens being passed in text chunks.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/tokenization/tokenizer.py
+++ b/curated_transformers/tokenization/tokenizer.py
@@ -1,5 +1,6 @@
 from typing import Iterable, List, Optional, Set, TypeVar, Union
 from abc import ABC, abstractmethod
+from ahocorasick import Automaton
 from dataclasses import dataclass
 import re
 import torch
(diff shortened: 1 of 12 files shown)


curated_transformers/tokenization/tokenizer.py
```diff
@@ -162,6 +171,43 @@ def _convert_strings(
             for seq in input
         ]
 
+    def _validate_chunks(self, input: List[InputChunks]):
+        if self._any_special_tokens_regex is None:
+            special_tokens = self._special_tokens()
+            if len(special_tokens) == 0:
+                return
+            pattern = "|".join(
```
danieldk, line 179: I think we should use Aho-Corasick here. Much faster than non-regular language 'regular expressions` implementation and no special escaping needed:

https://pypi.org/project/pyahocorasick/

shadeMe:
Closing this as discussed internally. We'll instead modify the tokenizer vocabs to not include special tokens/pieces.

PR #185, opened 2023-06-07
`StringGenerator`: decode all output ids at once

danieldk:
## Description

`Generator` returns an iterator, allowing the piece identifiers to be generated incrementally. Similarly, `StringGenerator` also returned an iterator, converting the piece identifiers to pieces. However, the string decoding used for this can fail since not every piece identifier represents a valid UTF-8 string by itself. For example, the byte BPE encoder uses bytes as the smallest unit, and a single byte can represent a UTF-8 initial or continuation byte. Attempting to convert such a byte to a string will be invalid.

Another issue is that potential decoding pre/post-processors may expect to see the full sequence.

This change addresses this shortcoming by accumulating the piece identifiers in `StringGenerator` and only decoding the piece identifiers once all sequences are done generating.

We might want to bring back incremental string generation in the future, but we need to extend the tokenizers to support this properly.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/curated_transformers/tests/generation/test_dolly_v2.py
+++ b/curated_transformers/tests/generation/test_dolly_v2.py
@@ -1,19 +1,9 @@
-from typing import Iterator, List, Tuple
 import pytest
 import torch
 
 from curated_transformers.generation.dolly_v2 import DollyV2Generator
 
 from ..conftest import GPU_TESTS_ENABLED
-
-
-def _collect_pieces(generator: Iterator[List[Tuple[int, str]]], n: int) -> List[str]:
-    pieces = [[] for _ in range(n)]
-    for step_pieces in generator:
-        for seq_id, piece in step_pieces:
-            pieces[seq_id].append(piece)
-
-    return ["".join(answer) for answer in pieces]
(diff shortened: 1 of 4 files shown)


curated_transformers/tests/generation/test_dolly_v2.py
```diff
@@ -27,7 +17,8 @@ def test_generate_deterministic():
         "What is the Rust programming language?",
         "What is spaCy?",
     ]
-    assert _collect_pieces(generator(prompts), len(prompts)) == [
+    print(generator(prompts))
```
shadeMe, line 20: `stdout` is captured by default, so this shouldn't affect `pytest` output, but I wanted to flag this just in case it's a debug statement.

shadeMe (approved):
I like the new, simpler API! 🎉

curated_transformers/tests/generation/test_dolly_v2.py
danieldk: Thanks! Yeah, these were debug statements, removed.

PR #186, opened 2023-06-08
`Generator`: add support for sampling

danieldk:
## Description

This change adds support for sampling to the generator. When sampling is used, we draw a random piece using the given softmax distribution. This adds more variety to the output than greedy/deterministic argmax generation.

To offer some initial control over the sampling, this change also adds top-k sampling and softmax temperature. These two techniques transform the logits before sampling. In order to allow such transforms, we introduce a new base class `LogitsTransform`, with derived `TopKTransform` and `TemperatureTransform` classes.

From the user's perspective, top-k and temperature are configured through the new `SampleGenerationConfig` class. For greedy generation, we add `GreedyGenerationConfig`.

## Checklist

- [ ] I confirm that I have the right to submit this contribution under the project's MIT license.

### DIFF
--- a/curated_transformers/generation/config.py
+++ b/curated_transformers/generation/config.py
@@ -51,7 +51,8 @@ class SampleGeneratorConfig(GeneratorConfig):
         - T < 1: the entropy of the distribution is decreased.
         - T > 1: the entropy of the distribution is increased.
     :param top_k:
-        Sample from top-k highest-probability pieces.
+        Sample from top-k highest-probability pieces. k < 1 disables top-k
+        filtering.
     """
 
     temperature: float = 1.0
--- a/curated_transformers/generation/config.py
+++ b/curated_transformers/generation/config.py
@@ -1,0 +1,66 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+from .logits import (
+    CompoundLogitTransforms,
+    LogitsTransform,
+    TemperatureTransform,
+    TopKTransform,
+)
+
+
+class GeneratorConfig(ABC):
+    """Configuration of the generator."""
+
+    @abstractmethod
+    def logits_transform(self) -> LogitsTransform:
+        """
+        Get logit transform for the configuration.
+
+        :returns:
+            Logits transform. Usually multiple composed transforms.
+        """
+        ...
+
+
+class GreedyGeneratorConfig(GeneratorConfig):
+    """
+    Configuration for greedy generation.
+
+    Greedy generation always selects the highest-probability piece, leading
+    to deterministic generation.
+    """
+
+    def logits_transform(self) -> LogitsTransform:
+        return CompoundLogitTransforms([])
+
+
+@dataclass
+class SampleGeneratorConfig(GeneratorConfig):
+    """
+    Configuration for generation with sampling.
+
+    Sampling-based generation samples pieces from probability distributions.
+    Generation is non-deterministic as a result, but provides more varied
+    output.
+
+    :param temperature:
+        Softmax temperature. For a temperature T:
+
+        - T = 1: the distribution is not changed.
+        - T < 1: the entropy of the distribution is decreased.
+        - T > 1: the entropy of the distribution is increased.
+    :param top_k:
+        Sample from top-k highest-probability pieces.
+    """
+
+    temperature: float = 1.0
+    top_k: int = 0
+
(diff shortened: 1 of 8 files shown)


curated_transformers/generation/logits.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class LogitsTransform(ABC):
+    """
+    A logits transform changes the logits of a softmax distribution in some
+    way. For instance, :class:`TemperatureTransform` changes the temperature
+    of the softmax distribution.
+    """
+
+    def __call__(self, logits: Tensor, inplace=False) -> Tensor:
```
shadeMe, line 14:
```suggestion
    def __call__(self, logits: Tensor, inplace: bool = False) -> Tensor:
```

curated_transformers/generation/logits.py
```diff
@@ -0,0 +1,118 @@
...
+        Transform the given logits.
+
+        :param logits:
+            The logits array.
+            **Shape:** (..., n_class)
+        :param inplace:
+            Transform logits in-place.
+        :returns:
+            The transformed logits. This will be the same tensor object as
+            the ``logits`` argument when ``inplace`` is set.
```
shadeMe, line 25: Are the double back-ticks intentional?

curated_transformers/generation/logits.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, k: int):
+        """
+        Construct a top-k logits transform.
+
+        :param k:
+            The value of k in top-k. The transform is a noop for values
+            less than 1.
+        """
+        super().__init__()
+        self.k = k
```
shadeMe, line 71: What speaks against making `k>=1` an requirement?

curated_transformers/generation/logits.py
```diff
@@ -0,0 +1,118 @@
...
+        p(y|x) = \frac{\mbox{exp}(\frac{z_i}{T})}{\sum_{y'=1}^N{\mbox{exp}(\frac{z_{y'}}{T})}}
+
+    For a temperature T:
+
+    - T = 1: the distribution is not changed.
+    - T < 1: the entropy of the distribution is decreased.
+    - T > 1: the entropy of the distribution is increased.
+    """
+
+    def __init__(self, temperature=1.0):
```
shadeMe, line 100:
```suggestion
    def __init__(self, temperature: float = 1.0):
```

curated_transformers/tests/generation/test_dolly_v2.py
```diff
@@ -26,7 +33,35 @@ def test_generate_deterministic():
...
+        "What is spaCy?",
+    ]
+    # Fix the seed so that we are always randomly sampling in the same way.
+    torch.manual_seed(0)
+    assert dolly_generator(prompts, config=SampleGeneratorConfig(top_k=10)) == [
+        "SpaCy (short for Spanish-Portuguese-Chinese artificial intelligence language model) is an open-source language model (LLM) that uses a novel combination of deep learning, big-data, and semantic parsing to provide the best accuracy available with no human involvement. It is one of the most effective open-source language models.\n\n",
+        "SpaCy is an open-source package for Python that provides language model (LM) and automatic speech recognition (ASR) functionality.\n\n",
+    ]
+
+    torch.manual_seed(0)
```
shadeMe, line 56: Do we need to set this multiple times?

curated_transformers/generation/logits.py
danieldk: You want some value of `k` where you can effectively disable it. It seems that other implementations use `k<1` to do this. You could set `k` to a large value, bet then you need to know the vocab size.

curated_transformers/generation/config.py
```diff
@@ -0,0 +1,66 @@
...
+    output.
+
+    :param temperature:
+        Softmax temperature. For a temperature T:
+
+        - T = 1: the distribution is not changed.
+        - T < 1: the entropy of the distribution is decreased.
+        - T > 1: the entropy of the distribution is increased.
+    :param top_k:
+        Sample from top-k highest-probability pieces.
```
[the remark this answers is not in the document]
danieldk: Fixed.

curated_transformers/tests/generation/test_dolly_v2.py
danieldk: Removed.

shadeMe (approved):

PR #187, opened 2023-06-12
Refactor model loading code in preparation of quantization support

shadeMe:
## Description

This PR refactors the model loading code to add support for the upcoming `bitsandbytes` quantization backend for mixed  `int8` and `float16` inference. 

Furthermore, it introduces the concept of deserialization buckets for models that implement the `FromPretrainedHF` mixin. The process of converting HF checkpoints to our format includes the merging of the query, key and value projection layers in the self-attention module into a single dense layer. This has the implicit pre-condition that the parameters of those layers are always processed at the same time. This assumption breaks down when loading from a sharded HF checkpoint as the constituent layers' parameters can potentially be spread across different shards.

To workaround this, we accumulate those parameters in per-transfomer-layer buckets and deserialize them at once. 

### Types of change

Refactor/Enhancement

## Checklist


curated_transformers/util/serde.py
```diff
@@ -0,0 +1,311 @@
...
+    :param parameter_name:
+        Name of the parameter being converted/replaced.
+    :param tensor:
+        Tensor to be converted.
+    :param device:
+        Device to which the converted parameter is moved.
+    :returns:
+        Converted parameter.
+    """
+    old_param = module._parameters[parameter_name]
```
danieldk, line 200: Can we use

https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.get_parameter

here rather than relying on internals?

curated_transformers/util/pytorch.py
```diff
@@ -0,0 +1,36 @@
...
+    :param func:
+        A callable that takes a module iterator as its argument.
+    """
+    queue: List[ModuleIterator] = [ModuleIterator(module, "", "", None)]
+
+    while queue:
+        itr = queue.pop(0)
+        func(itr)
+
+        for name, child in itr.module._modules.items():
```
danieldk, line 33: Can we use `named_children` here?

https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_children

curated_transformers/models/albert/_hf.py
```diff
@@ -74,29 +75,42 @@ def convert_hf_state_dict(params: Mapping[str, Parameter]) -> Mapping[str, Tenso
...
-    out["embeddings.layer_norm.bias"] = stripped_params["embeddings.LayerNorm.bias"]
-
-    # Embedding projection
-    out["embeddings.projection.weight"] = stripped_params[
-        "encoder.embedding_hidden_mapping_in.weight"
-    ]
-    out["embeddings.projection.bias"] = stripped_params[
-        "encoder.embedding_hidden_mapping_in.bias"
-    ]
+    key_map = {
```
danieldk, line 78: Now that these are nicely bundled, it's probably better so make a top-level `KEY_MAP` constant. We could probably use

https://docs.python.org/3/library/types.html#types.MappingProxyType

if we are worried about external modification?

danieldk (commented):
Nice work! 🎉

curated_transformers/util/serde.py
shadeMe: We could, but we'd need to reconstruct the fully-qualified key using both `module_prefix` and `parameter_name` to pass to `get_parameter`. Also, we still have to manually modify the `_parameters` dict when we replace the old parameter with the new one, so the reliance on internals will still be there.

curated_transformers/util/serde.py
```diff
@@ -0,0 +1,311 @@
...
+        if len(self.expected_keys) != len(self):
+            return False
+
+        seen = 0
+        for expected in self.expected_keys:
+            for key in self.keys():
+                if expected in key:
+                    seen += 1
+                    break
+        return seen == len(self.expected_keys)
```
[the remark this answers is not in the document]
shadeMe: This won't work as such since the `expected_key` strings don't contain the full key for the parameter.

curated_transformers/util/pytorch.py
```diff
@@ -0,0 +1,36 @@
+from typing import Callable, List, Optional
+from dataclasses import dataclass
+from torch.nn import Module
+
+
+@dataclass
+class ModuleIterator:
```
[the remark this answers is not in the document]
shadeMe: Fixed.

curated_transformers/util/serde.py
```diff
@@ -0,0 +1,311 @@
...
+        return True
+
+
+class RegExParameterBucket(DeserializationParamBucket):
+    """Groups parameters whose keys match a given regular expression."""
+
+    key_matcher: re.Pattern
+    expected_keys: Set[str]
+
+    def __init__(self, pattern: str, expected_keys: Set[str]) -> None:
```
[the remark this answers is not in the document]
shadeMe: Good point, changed.

curated_transformers/util/serde.py
```diff
@@ -0,0 +1,311 @@
...
+]
+
+# Args: State dict.
+# Returns the converted state dict.
+HFStateDictConverterT = Callable[
+    [Mapping[str, torch.Tensor]], Mapping[str, torch.Tensor]
+]
+
+
+class DeserializationParamBucket(UserDict):
```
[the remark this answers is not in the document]
shadeMe: Fixed.

curated_transformers/util/pytorch.py
shadeMe: Done.

curated_transformers/util/serde.py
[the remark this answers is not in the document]
danieldk: Ah, right!

curated_transformers/util/serde.py
danieldk: Ok, let's leave it like this, since it's to have the bucketing code in git.

PR #190, opened 2023-06-13
Customizable handling of query, key and value projections

shadeMe:
## Description

Introduces the `QkvMode` enumeration that can be used to control how the `SelfAttention` modules will handle the projection of the query, key and value tensors. This parameter is meant to be set on a per-model architectures basis.

As a side-effect of this change, we no longer need to perform the merging of the above tensors for all models, so we can eliminate the custom merging code. Furthermore, it also obviates the need for deserialization buckets since there are no inter-parameter dependencies during the HF parameter conversion process. Should we choose to make this parameter user-configurable in the future, we can still reintroduce the bucketing during loading.

### Types of change

Refactor/Enhancement

## Checklist

### DIFF
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -120,9 +120,9 @@ class QkvMode(IntEnum):
     """Modes of handling the query, key and value projections
     in the self-attention layer.
 
-    ``SEPARATE`` - Use separate projections for query, key and value.
-    ``MERGED_SPLIT_BEFORE`` - Use a merged projection for query, key and value, and split heads before chunking.
-    ``MERGED_SPLIT_AFTER`` - Use a merged projection for query, key and value, and split heads after chunking.
+    - ``SEPARATE`` - Use separate projections for query, key and value.
+    - ``MERGED_SPLIT_BEFORE`` - Use a merged projection for query, key and value, and split heads before splitting the query, key and value representations.
+    - ``MERGED_SPLIT_AFTER`` - Use a merged projection for query, key and value, and split heads after splitting the query, key and value representations.
     """
 
     SEPARATE = (0,)
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -203,10 +203,10 @@ class SelfAttention(Module):
     def __init__(
         self,
         *,
-        dropout_prob: float = 0.1,
-        hidden_width: int = 768,
-        num_attention_heads: int = 12,
-        qkv_mode: QkvMode = QkvMode.SEPARATE,
+        dropout_prob: float,
+        hidden_width: int,
+        num_attention_heads: int,
+        qkv_mode: QkvMode,
         device: Optional[torch.device] = None,
     ):
         """Construct a self-attention layer.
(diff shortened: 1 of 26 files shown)


curated_transformers/models/attention.py
```diff
@@ -191,6 +206,7 @@ def __init__(
         dropout_prob: float = 0.1,
         hidden_width: int = 768,
         num_attention_heads: int = 12,
+        qkv_mode: QkvMode = QkvMode.SEPARATE,
```
danieldk, line 209: I wouldn't specify a default value for this argument, this is something we want the implementation to always explicitly specify. The same is probably true for the `dropout_prob`, `hidden_width` and `num_attention_heads` arguments, where there is no good default.

curated_transformers/models/attention.py
```diff
@@ -269,9 +305,9 @@ def __init__(
         dropout_prob: float = 0.1,
         hidden_width: int = 768,
         num_attention_heads: int = 12,
+        qkv_mode: QkvMode = QkvMode.MERGED_SPLIT_AFTER,
```
danieldk, line 308: Same here.

curated_transformers/models/bert/config.py
```diff
@@ -36,17 +38,20 @@ class BertAttentionConfig:
     hidden_width: int
     num_attention_heads: int
     dropout_prob: float
+    qkv_mode: QkvMode
```
danieldk, line 41: I don't think we should at this to the attention config until a need arises. This is really a property of the architecture and we should just hardcode it where the attention layer is constructed.

curated_transformers/models/attention.py
```diff
@@ -115,6 +116,20 @@ def filter_batch_items(self, mask: Tensor) -> "KeyValueCache":
         return KeyValueCache(key=self.key[mask], value=self.value[mask])
 
 
+class QkvMode(IntEnum):
+    """Modes of handling the query, key and value projections
+    in the self-attention layer.
+
+    ``SEPARATE`` - Use separate projections for query, key and value.
+    ``MERGED_SPLIT_BEFORE`` - Use a merged projection for query, key and value, and split heads before chunking.
+    ``MERGED_SPLIT_AFTER`` - Use a merged projection for query, key and value, and split heads after chunking.
```
danieldk, line 125: I wonder how this is rendered once we expose this in the docs? Aren't these lines rendered as a single paragraph, maybe it's worth prefixing it by `- ` to get a bulleted list. Or perhaps the `#: ` here on the fields, since they are not constructor arguments here?

danieldk, line 125: I took the term _chunking_ from the `chunk` op, but maybe it's not clear what this means? Perhaps something like:

_Use a merged projection for query, key and value, and split heads before splitting the query, key and value representations._

?

curated_transformers/models/attention.py
danieldk: But this isn't `BertAttentionConfig`, but the constructor of the generic attention layer.

shadeMe: Whoops! I didn't notice that :sweat_smile: Yeah, that makes sense :+1:

shadeMe: Done.

curated_transformers/models/bert/config.py
shadeMe: Removed.

curated_transformers/models/attention.py
shadeMe: Oh yeah, it does appear as a single paragraph. Moved the string to the variants.

shadeMe: Done.

danieldk (approved):

PR #193, opened 2023-06-15
Clean up the self-attention layer

danieldk:
## Description

- Remove SelfAttention, rename SelfAttentionWithRotaryEmbeddings to SelfAttention.
- Factor out application of rotary embeddings to query and key to a separate module, to clean up SelfAttention.
- Factor out query, key, value and head construction, splitting, and head splitting to a separate method.
- Move cache protocol out of the attention module to avoid cycles.
- Rename mha_layer_norm to input_layer_norm, since in models with parallel connections this is fed both to both the attention block and the ff layers.
- Swap key, query argument order of ScaledDotProductAttention to be more canonical. Also make kwarg-only to avoid passing tensors in the wrong order.

### Types of change

Maintenance

## Checklist

### DIFF
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -174,6 +174,8 @@ class SelfAttention(Module):
     Transformer self-attention layer (Vaswani et al., 2017).
     """
 
+    rotary_embeds: Optional[QueryKeyRotaryEmbeddings]
+
     def __init__(
         self,
         *,
@@ -215,6 +217,8 @@ class SelfAttention(Module):
                 fraction=rotary_embeds.fraction,
                 dims_per_head=self.dims_per_head,
             )
+        else:
+            self.rotary_embeds = None
 
         self.attention = ScaledDotProductAttention(
             dropout_prob=dropout_prob,
@@ -271,7 +275,7 @@ class SelfAttention(Module):
 
         query, key, value = self._query_key_value(x)
 
-        if hasattr(self, "rotary_embeds"):
+        if self.rotary_embeds is not None:
             query, key = self.rotary_embeds(
                 query=query, key=key, cache=cache, positions=positions
             )
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -1,13 +1,14 @@
 import math
 from dataclasses import dataclass
 from enum import IntEnum
-from typing import Optional, Protocol, Tuple, TypeVar
+from typing import Optional, Tuple
 
 import torch
 from torch import Tensor
 from torch.nn import Linear, Module
 
-from .embeddings import RotaryEmbeddings
+from .embeddings import QueryKeyRotaryEmbeddings
+from .output import KeyValueCache
 
 
 class AttentionMask:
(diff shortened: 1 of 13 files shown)


curated_transformers/models/attention.py
```diff
@@ -318,9 +191,8 @@ def __init__(
         :param hidden_width: Hidden width of the layer.
         :param num_attention_heads: Number of attention heads.
         :param qkv_mode: Handling mode for query, key and value.
-        :param rotary_fraction: fraction of hidden width to apply rotary
-            embeddings to. Must be in [0,1].
-        :param rotary_base: Base in signifying the rotary embedding period.
+        :param rotary_embeds: Configuration for rotary embeddings, rotary
+            embeddings wil not be used when set to ``None``.
```
shadeMe, line 195:
```suggestion
            embeddings will not be used when set to ``None``.
```

curated_transformers/models/attention.py
```diff
@@ -307,8 +181,7 @@ def __init__(
         hidden_width: int,
         num_attention_heads: int,
         qkv_mode: QkvMode,
-        rotary_fraction: float,
-        rotary_base: int = 10000,
+        rotary_embeds: Optional[RotaryEmbeddingConfig],
```
shadeMe, line 184:
```suggestion
        rotary_embeds: Optional[RotaryEmbeddingConfig] = None,
```
Do we want to default to `None` here as well?

curated_transformers/models/attention.py
```diff
@@ -393,74 +269,77 @@ def forward(
...
-                    device=k_rotary.device,
-                ).repeat(x.size(0), 1)
-
-            # Apply rotary embeddings.
-            k_rotary = self.rotary_embeds(k_rotary, positions=positions)
-            q_rotary = self.rotary_embeds(q_rotary, positions=positions)
-
-            q = torch.cat([q_rotary, q_rest], dim=-1)
-            k = torch.cat([k_rotary, k_rest], dim=-1)
+        if hasattr(self, "rotary_embeds"):
```
shadeMe, line 274: I think we can set `self.rotary_embeds` to `None` if the config is not provided and use a `None` check here.

curated_transformers/models/embeddings.py
```diff
@@ -155,3 +157,99 @@ def forward(self, input: torch.Tensor, *, positions: Optional[Tensor] = None):
...
+        fraction: float,
+        dims_per_head: int,
+        device: Optional[torch.device] = None,
+    ) -> None:
+        """
+        Construct rotary embeddings for transforming key and value
+        representations.
+
+        :param base: Base in signifying the rotary embedding period.
+        :param fraction: fraction of hidden width to apply rotary
```
shadeMe, line 178:
```suggestion
        :param fraction: Fraction of hidden width to apply rotary
```

curated_transformers/models/attention.py
danieldk: Yeah, I was doubting what to do, since the constructor is a little longer when making it an `Optional`, but it is probably clearer, so changed.

shadeMe (approved):

PR #194, opened 2023-06-16
Add FalconGenerator and all the needed machinery

danieldk:
## Description

This change adds support for generation using the Falcon models. The generator wrapper is FalconGenerator. Falcon uses the Refined Web Model, which is slightly different from other models that we currently have:
    
- The model can use shared key and value heads. This is implemented using the new qkv_head_sharing argument of the self attention layer along with the QkvHeadSharing enum.
- The model can use all linear layers without bias. This is accommodated by adding a new use_bias argument to the self attention layer and the pointwise feed-forward layer.
- The activation is always GELU.
    
Unfortunately, we cannot test decoding with cache, position tensors, or the causal LM against the upstream implementation, since it is broken in two ways:
    
- The upstream implementation uses the PyTorch scaled dot product attention, which creates invalid causal attention masks. When the query and key shapes differ.
- The upstream implementation does not take into account that rotary embeddings should be offset for position when using a cache. As a result, the new query and key pieces will always be treated as starting at position 0.
    
I still need to investigate the differences between the NeoX tokenizer and the Refined Web Model tokenizer. I'll address this in a future PR.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -73,6 +73,16 @@ def apply_causal_mask(input: Tensor) -> Tensor:
     return torch.where(causal_mask, input, blocked_value)
 
 
+class QkvHeadSharing(IntEnum):
+    """Sharing of head parameters."""
+
+    #: No parameters are shared between heads.
+    NONE = 0
+
+    #: Key shares heads, value shares heads, query has separate heads.
+    KEY_VALUE = 1
+
+
 class QkvMode(IntEnum):
     """Modes of handling the query, key and value projections
     in the self-attention layer.
@@ -180,16 +190,19 @@ class SelfAttention(Module):
         self,
         *,
         dropout_prob: float,
+        qkv_head_sharing: QkvHeadSharing,
         hidden_width: int,
         num_attention_heads: int,
         qkv_mode: QkvMode,
         rotary_embeds: Optional[RotaryEmbeddingConfig] = None,
+        use_bias: bool,
         device: Optional[torch.device] = None,
     ):
         """Construct a self-attention layer with rotary position embeddings.
 
         :param dropout_prob: Dropout to apply between the self-attention
             and output layers.
+        :param qkv_head_sharing: Head sharing in query, key and value.
         :param hidden_width: Hidden width of the layer.
         :param num_attention_heads: Number of attention heads.
         :param qkv_mode: Handling mode for query, key and value.
@@ -224,18 +237,40 @@ class SelfAttention(Module):
             dropout_prob=dropout_prob,
         )
 
+        if (
+            qkv_mode == QkvMode.MERGED_SPLIT_BEFORE
+            and qkv_head_sharing == QkvHeadSharing.KEY_VALUE
+        ):
+            raise ValueError(
+                "QkvMode.MERGED_SPLIT_BEFORE is incompatible with key/value head sharing"
+            )
+        self.head_sharing = qkv_head_sharing
+
+        # Head sharing is implemented as just having one head. We will still
+        # have multiple heads after attention, because the value has multiple
+        # heads, so attention will broadcast.
+        kv_dim = (
+            self.dims_per_head
+            if qkv_head_sharing == QkvHeadSharing.KEY_VALUE
+            else self.model_dim
+        )
         if qkv_mode == QkvMode.SEPARATE:
-            self.query = Linear(self.model_dim, self.model_dim, device=device)
-            self.key = Linear(self.model_dim, self.model_dim, device=device)
-            self.value = Linear(self.model_dim, self.model_dim, device=device)
+            self.query = Linear(
+                self.model_dim, self.model_dim, bias=use_bias, device=device
+            )
+            self.key = Linear(self.model_dim, kv_dim, bias=use_bias, device=device)
+            self.value = Linear(self.model_dim, kv_dim, bias=use_bias, device=device)
         else:
             self.input = Linear(
                 self.model_dim,
-                self.model_dim * 3,
+                self.model_dim + 2 * kv_dim,
+                bias=use_bias,
                 device=device,
             )
 
-        self.output = Linear(self.model_dim, self.model_dim, device=device)
+        self.output = Linear(
+            self.model_dim, self.model_dim, bias=use_bias, device=device
+        )
 
     def forward(
         self,
@@ -315,14 +350,17 @@ class SelfAttention(Module):
             Query, key, value
             **Shape:** (batch, head, seq_len, width_per_head)
         """
+        kv_heads = (
+            1 if self.head_sharing == QkvHeadSharing.KEY_VALUE else self.num_heads
+        )
         if self.qkv_mode == QkvMode.SEPARATE:
             query = self.query(x)
             key = self.key(x)
             value = self.value(x)
 
             query = split_heads(query, self.num_heads)
-            key = split_heads(key, self.num_heads)
-            value = split_heads(value, self.num_heads)
+            key = split_heads(key, kv_heads)
+            value = split_heads(value, kv_heads)
         elif self.qkv_mode == QkvMode.MERGED_SPLIT_BEFORE:
             proj = self.input(x)
             proj = split_heads(proj, self.num_heads)
@@ -332,15 +370,15 @@ class SelfAttention(Module):
             query, key, value = proj.split(
                 [
                     self.num_heads * self.dims_per_head,
-                    self.num_heads * self.dims_per_head,
-                    self.num_heads * self.dims_per_head,
+                    kv_heads * self.dims_per_head,
+                    kv_heads * self.dims_per_head,
                 ],
                 dim=-1,
             )
 
             query = split_heads(query, self.num_heads)
-            key = split_heads(key, self.num_heads)
-            value = split_heads(value, self.num_heads)
+            key = split_heads(key, kv_heads)
+            value = split_heads(value, kv_heads)
 
         return query, key, value
 
--- a/curated_transformers/models/attention.py
+++ b/curated_transformers/models/attention.py
@@ -80,6 +80,7 @@ class QkvHeadSharing(IntEnum):
     NONE = 0
 
     #: Key shares heads, value shares heads, query has separate heads.
+    #: See Shazeer, 2019: https://arxiv.org/abs/1911.02150
     KEY_VALUE = 1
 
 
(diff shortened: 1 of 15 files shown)


curated_transformers/models/attention.py
```diff
@@ -73,6 +73,16 @@ def apply_causal_mask(input: Tensor) -> Tensor:
     return torch.where(causal_mask, input, blocked_value)
 
 
+class QkvHeadSharing(IntEnum):
+    """Sharing of head parameters."""
+
+    #: No parameters are shared between heads.
+    NONE = 0
+
+    #: Key shares heads, value shares heads, query has separate heads.
```
shadeMe, line 82:
```suggestion
    #: Key and value share heads, query has separate heads.
```

curated_transformers/models/attention.py
```diff
@@ -224,18 +237,40 @@ def __init__(
...
+                self.model_dim, self.model_dim, bias=use_bias, device=device
+            )
+            self.key = Linear(self.model_dim, kv_dim, bias=use_bias, device=device)
+            self.value = Linear(self.model_dim, kv_dim, bias=use_bias, device=device)
         else:
             self.input = Linear(
                 self.model_dim,
-                self.model_dim * 3,
+                self.model_dim + 2 * kv_dim,
+                bias=use_bias,
```
shadeMe, line 267: Was the absence of `bias` an oversight until now or is it required for Falcon?

curated_transformers/models/feedforward.py
```diff
@@ -31,12 +31,17 @@ def __init__(
         hidden_act: str = "gelu",
         hidden_width: int = 768,
         intermediate_width: int = 3072,
+        use_bias: bool,
```
shadeMe, line 34: We don't want a default here?

curated_transformers/models/attention.py
```diff
@@ -73,6 +73,16 @@ def apply_causal_mask(input: Tensor) -> Tensor:
...
 
 
+class QkvHeadSharing(IntEnum):
+    """Sharing of head parameters."""
+
+    #: No parameters are shared between heads.
+    NONE = 0
+
+    #: Key shares heads, value shares heads, query has separate heads.
+    KEY_VALUE = 1
```
shadeMe, line 83: We could also mention [the paper](https://arxiv.org/abs/1911.02150) here.

curated_transformers/models/refined_web_model/causal_lm.py
```diff
@@ -0,0 +1,75 @@
...
+        :param config: Model configuration.
+        :param device: Device on which the module is to be initialized.
+        """
+        super().__init__()
+
+        self.decoder = RefinedWebModelDecoder(config, device=device)
+        self.output_embeddings = Linear(
+            in_features=config.layer.hidden_width,
+            out_features=config.embedding.vocab_size,
+            bias=False,
```
shadeMe, line 35: No `bias` term for the LM head (because of the layernorm that's performed on the last layer's output)?

curated_transformers/models/refined_web_model/config.py
```diff
@@ -0,0 +1,175 @@
...
+        use_bias: bool = False,
+    ):
+        """
+        :param dropout_prob: Dropout to apply after attention.
+        :param hidden_width: Hidden width of the transformer.
+        :param multi_query: Use multiple query heads and single key and
+            value heads.
+        :param num_attention_heads: Number of attention heads.
+        :param rotary_base: Base in signifying the rotary embedding period.
+        :param rotary_fraction: fraction of hidden width to apply rotary
```
shadeMe, line 34:
```suggestion
        :param rotary_fraction: Fraction of hidden width to apply rotary
```

curated_transformers/models/refined_web_model/config.py
```diff
@@ -0,0 +1,175 @@
...
+            embeddings to. Must be in [0,1].
+        :param use_bias: use bias in linear layers.
```
shadeMe, line 36:
```suggestion
        :param use_bias: Use bias in linear layers.
```

curated_transformers/models/refined_web_model/config.py
```diff
@@ -0,0 +1,175 @@
...
+        num_hidden_layers: int = 32,
+        layer_norm_eps: float = 1e-5,
+        use_bias: bool = False,
+    ) -> None:
+        """
+        :param dropout_prob: Dropout to apply after hidden layers.
+        :param hidden_width: Hidden width of the transformer.
+        :param layer_norm_eps: Epsilon for layer normalization.
+        :param num_hidden_layers: Number of hidden layers.
+        :param use_bias: use bias in linear layers.
```
shadeMe, line 102:
```suggestion
        :param use_bias: Use bias in linear layers.
```

curated_transformers/models/refined_web_model/config.py
```diff
@@ -0,0 +1,175 @@
...
+            embedding layers.
+        :param hidden_width: Hidden width of the transformer.
+        :param layer_norm_eps: Epsilon for layer normalization.
+        :param multi_query: Use multiple query heads and single key and
+            value heads.
+        :param num_hidden_layers: Number of hidden layers.
+        :param rotary_embedding_base: Base in signifying the rotary embedding period.
+        :param rotary_embedding_fraction: Fraction of hidden width to apply rotary
+            embeddings to. Must be in [0,1].
+        :param use_bias: use bias in linear layers.
```
shadeMe, line 146:
```suggestion
        :param use_bias: Use bias in linear layers.
```

curated_transformers/models/refined_web_model/layer.py
```diff
@@ -0,0 +1,108 @@
...
+    def forward(
+        self,
+        x: Tensor,
+        attention_mask: Optional[AttentionMask],
+        cache: Optional[KeyValueCache] = None,
+        positions: Optional[Tensor] = None,
+        store_cache: bool = False,
+    ) -> Tuple[Tensor, Optional[KeyValueCache]]:
+        """
+        Apply the GPT-NeoX layer to the given piece hidden representations.
```
shadeMe, line 74:
```suggestion
        Apply the Refined Web Model layer to the given piece hidden representations.
```

curated_transformers/models/refined_web_model/layer.py
```diff
@@ -0,0 +1,108 @@
...
+            look up rotary embeddings. Normally, these positions are calculated
+            automatically. But if the positions deviate for some reason, they
+            can be provided through this argument.
+        :param store_cache: Whether to cache the key/value representations for
+            future reuse.
+        :returns: Layer output.
+
+        Shapes:
+            x - (batch, seq_len, width)
+            attention_mask - (batch, seq_len)
```
shadeMe, line 92: We should move the shape definitions to each parameter's docstring.

curated_transformers/tests/generation/test_falcon.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.mark.veryslow
+@pytest.mark.skipif(not GPU_TESTS_ENABLED, reason="requires GPU")
+def test_generate_deterministic(falcon_generator):
+    prompts = [
+        "What is the Rust programming language?",
+        "What is spaCy?",
+    ]
+    assert falcon_generator(prompts, config=GreedyGeneratorConfig()) == [
```
shadeMe, line 27: The test will look less noisy (and more debuggable) if we extracted the expected outputs to a variable and compared the outputs with it (`expected = expected[::-1]` for the reversed version).

curated_transformers/models/attention.py
danieldk: `bias` is enabled by default. Falcon is the first model that we support that doesn't use bias for linear layers.

curated_transformers/models/feedforward.py
danieldk: I think we should do the opposite here and remove all the defaults (like attention). I'll add a commit to do this.

curated_transformers/models/refined_web_model/causal_lm.py
danieldk: It's common to not have bias on the output embeddings on the output. I don't really like the LM head terminology, it's really more like: the model produces an embedding and you then do a matrix multiplication between the predicted embedding and the output embedding matrix to find the token for which the output embedding is the closest to the predicted embedding.

curated_transformers/models/attention.py
danieldk: I intentionally formulated it like this, otherwise it reads like the key and the value literally use the same head (v = k).

curated_transformers/models/feedforward.py
danieldk: Removed all the defaults.

curated_transformers/models/attention.py
shadeMe: Ah, I see.

curated_transformers/tests/generation/test_falcon.py
danieldk: That's a great idea! Changed the Dolly tests in the same way.

curated_transformers/models/refined_web_model/layer.py
danieldk: Done. Also in the decoder module.

curated_transformers/models/attention.py
danieldk: Added.

curated_transformers/models/feedforward.py
```diff
@@ -16,24 +16,24 @@ class PointwiseFeedForward(Module):
...
-        hidden_width: int = 768,
-        intermediate_width: int = 3072,
+        hidden_act: str,
+        hidden_width: int,
+        intermediate_width: int,
         use_bias: bool,
         device: Optional[torch.device] = None,
     ):
+        """
+        :param hidden_act: the activation function to apply, one of: "relu",
```
shadeMe, line 31:
```suggestion
        :param hidden_act: The activation function to apply, one of: "relu",
```

curated_transformers/models/feedforward.py
```diff
@@ -16,24 +16,24 @@ class PointwiseFeedForward(Module):
...
+            "gelu" or "gelu_new".
+        :param hidden_width: the input and output width of the layer.
```
shadeMe, line 33:
```suggestion
        :param hidden_width: The input and output width of the layer.
```

curated_transformers/models/feedforward.py
```diff
@@ -16,24 +16,24 @@ class PointwiseFeedForward(Module):
...
+        :param intermediate_width: the width of the projection to which the
```
shadeMe, line 34:
```suggestion
        :param intermediate_width: The width of the projection to which the
```

shadeMe (approved):
LGTM! Just some minor typos.

PR #199, opened 2023-06-20
Add `AutoModel` classes + Refactoring

shadeMe:
## Description

Adds `AutoEncoder`, `AutoDecoder`, `AutoCausalLM` and `AutoGenerator` classes for automatically choosing the right internal component class. 

Plus some sundry refactoring and fixes.

### Types of change

Enhancement

## Checklist


curated_transformers/util/auto_model.py
```diff
@@ -0,0 +1,227 @@
...
+class AutoGenerator(AutoModel[GeneratorWrapper]):
+    """Causal LM generator loaded from the Hugging Face Model Hub.
+
+    **NOTE** - This class can currently only be used with following models:
+        - ``databricks/dolly-v2`` variants.
+        - ``tiiuae/falcon`` variants.
+    """
+
+    _DOLLY_V2_PREFIX = "databricks/dolly-v2"
+    _FALCON_PREFIX = "tiiuae/falcon"
```
danieldk, line 157: I think we should match for `dolly-v2` and `falcon` occurring in the string. Usually there are a lot of custom finetuned flavors and it would be a shame if they can't be used.

curated_transformers/util/auto_model.py
```diff
@@ -0,0 +1,227 @@
...
+        return generator
+
+
+def _instantiate_module_from_hf_hub(
+    name: str,
+    revision: str,
+    device: Optional[torch.device],
+    quantization_config: Optional[BitsAndBytesConfig],
+    model_type_to_class_map: Mapping[str, Type],
+    module_type: str,
```
danieldk, line 204: I am not a big fan of this stringly-typed argument. Maybe we could make this a class method of `AutoModel`, so that we could say `AutoDecoder`/`AutoEncoder`/`AutoCausalLM` below?

curated_transformers/tests/models/test_auto_models.py
```diff
@@ -0,0 +1,86 @@
...
+)
+from curated_transformers.util.auto_model import (
+    AutoCausalLM,
+    AutoDecoder,
+    AutoEncoder,
+    AutoGenerator,
+)
+
+
+@pytest.mark.skipif(not has_hf_transformers, reason="requires huggingface transformers")
```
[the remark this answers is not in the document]
shadeMe: Good point!

curated_transformers/util/auto_model.py
shadeMe: Yeah, I agree that it's iffy. Will add a class method.

danieldk (approved):
There are some merge conflicts, but looks good to me otherwise.

PR #200, opened 2023-06-20
Add LLaMA decoder and causal LM

danieldk:
## Description

LLaMA is pretty similar to NeoX, with some small modifications:

- Gated Linear Units
- RMS norm rather than layer norm
- No parallel attention

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/models/feedforward.py
+++ b/curated_transformers/models/feedforward.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Callable, Dict, Optional
 
 import torch
 from torch import Tensor
@@ -6,9 +6,22 @@ from torch.nn import Linear, Module
 
 from .activations import GeluFast, GeluNew
 
+_ACTIVATIONS: Dict[str, Callable[[], Module]] = {
+    "relu": torch.nn.ReLU,
+    "gelu": torch.nn.GELU,
+    "gelu_fast": GeluFast,
+    # Ideally, we would use torch.nn.GELU(approximate="tanh"). However,
+    # the differences between that and the manual Torch implementation
+    # are large enough to fail tests comparing output to HF
+    # transformers.
+    "gelu_new": GeluNew,
+    "silu": torch.nn.SiLU,
+}
+
 
 class PointwiseFeedForward(Module):
-    """Point-wise feed-forward layer (Vaswani et al., 2017).
+    """
+    Point-wise feed-forward layer (Vaswani et al., 2017).
 
     This layer is applied pointwise, meaning that the same
     transformation is applied to each sequence element. This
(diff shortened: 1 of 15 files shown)


curated_transformers/models/feedforward.py
```diff
@@ -1,30 +1,51 @@
-from typing import Optional
+from typing import Callable, Dict, Optional
 
 import torch
 from torch import Tensor
 from torch.nn import Linear, Module
 
 from .activations import GeluFast, GeluNew
 
+_ACTIVATIONS: Dict[str, Callable[[], Module]] = {
```
shadeMe, line 9: Can't we use `Type` instead of `Callable[...]`?

curated_transformers/models/llama/config.py
```diff
@@ -0,0 +1,174 @@
...
+        num_attention_heads: int = 32,
+        rotary_base=10000,
+        rotary_fraction=0.25,
+    ):
+        """
+        :param dropout_prob: Dropout to apply after attention.
+        :param hidden_width: Hidden width of the transformer.
+        :param num_attention_heads: Number of attention heads.
+        :param rotary_base: Base in signifying the rotary embedding period.
+        :param rotary_fraction: fraction of hidden width to apply rotary
```
shadeMe, line 31:
```suggestion
        :param rotary_fraction: Fraction of hidden width to apply rotary
```

curated_transformers/models/llama/decoder.py
```diff
@@ -0,0 +1,121 @@
...
+        :param positions: Input positions. Positions are needed to
+            look up rotary embeddings. Normally, these positions are calculated
+            automatically. But if the positions deviate for some reason, they
+            can be provided through this argument.
+        :param store_cache: Whether to cache the key/value representations for
+            future reuse.
+        :returns: Decoder representations of the given pieces.
+
+        Shapes:
+            input_ids, attention_mask, positions - (batch, seq_len)
```
shadeMe, line 75: Shapes as part of param docstrings.

curated_transformers/models/normalization.py
```diff
@@ -0,0 +1,44 @@
...
+        Apply RMS normalization to a tensor.
+
+        :param input:
+            The tensor to apply normalization to.
+        :returns:
+            Normalized tensor.
+        """
+        # Zhang & Sennrich, Equation 4
+        rms = (
+            input.to(torch.float32)
```
shadeMe, line 37: Maybe a comment to clarify why we're doing the calculation in f32?

curated_transformers/models/feedforward.py
danieldk: Sounds good, fixed.

curated_transformers/models/llama/config.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@dataclass
+class LLaMAAttentionConfig:
+    """LLaMA (Touvron et al., 2023) attention configuration."""
+
+    dropout_prob: float
+    hidden_width: int
+    num_attention_heads: int
+    qkv_mode: QkvMode
```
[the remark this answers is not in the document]
danieldk: Removed, also in NeoX, where this was originally copied from.

curated_transformers/models/llama/decoder.py
danieldk: Fixed, also in NeoX, where this was originally copied from.

curated_transformers/models/normalization.py
danieldk: Added.

PR #203, opened 2023-06-22
Add a small generator example script

danieldk:
## Description

This example reads prompts as a JSON list and returns JSON list with the generated text.

### Types of change

Example

## Checklist

### DIFF
--- a/examples/generate.py
+++ b/examples/generate.py
@@ -14,14 +14,18 @@ from curated_transformers.generation.config import (
 from curated_transformers.models.attention import enable_torch_sdp
 from curated_transformers.util.auto_model import AutoGenerator
 
+EPILOG = """This program takes a JSON list of strings from the standard inputs as prompts. For example:
+
+["What is spaCy?", "What is Rust?"]
+
+The output is a JSON list with the following format:
+
+[{"prompt": "What is spaCy?", "answer": "..."}, {"prompt": "What is Rust?", "answer": "..."}]'"""
+
 parser = argparse.ArgumentParser(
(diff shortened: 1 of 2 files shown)


examples/generate.py
```diff
@@ -0,0 +1,98 @@
...
+parser.add_argument(
+    "--sample",
+    "-s",
+    action="store_true",
+    help="enable sampling",
+)
+parser.add_argument(
+    "--torch-sdp",
+    action="store_true",
+    help="enable Torch scaled dot product implementation",
```
shadeMe, line 59:
```suggestion
    help="enable Torch scaled dot product attention implementation",
```

examples/generate.py
```diff
@@ -0,0 +1,98 @@
...
+
+from curated_transformers.generation.config import (
+    GreedyGeneratorConfig,
+    SampleGeneratorConfig,
+)
+from curated_transformers.models.attention import enable_torch_sdp
+from curated_transformers.util.auto_model import AutoGenerator
+
+parser = argparse.ArgumentParser(
+    description="Generate from a list of prompts in JSON format",
```
shadeMe, line 18: It'd be good to also include an example of the format.

danieldk: But where? 🤔

shadeMe: I was thinking we could add it to the description string? Doesn't it get shown in the help text?

danieldk: Added an epilogue.

shadeMe (approved):

PR #206, opened 2023-06-23
Remove HF cache from CI + Remove `slow` test marker

shadeMe:
## Description

Since we've now switched to small testing models for our tests, we no longer need to worry about the overhead of downloading them from the Hugging Face Hub during CI. So, we can remove the `cache` action from CI (which was incidentally making things harder to test when one needed to update a cached model's config).

Also, we no longer use the `--slow` `pytest` marker since switching to the `--veryslow` marker for LLM tests. This was also removed.

### Types of change

CI

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,13 +40,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.8", "3.11"]
-        include:
-          - os: ubuntu-latest
-            hf-path: ~/.cache/huggingface/hub
-          - os: macos-latest
-            hf-path: ~/.cache/huggingface/hub
-          # - os: windows-latest
-          #   hf-path: $HOME\\.cache\\huggingface\\hub
     steps:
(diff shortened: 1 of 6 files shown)


shadeMe:
GPU tests will fail until [this PR](https://github.com/explosion/buildkite-pipelines/pull/32) is merged.

.github/workflows/test.yml
```diff
@@ -92,29 +85,17 @@ jobs:
           python -m pip install -r requirements.txt
 
       - name: Run pytest
-        run: python -m pytest --pyargs curated_transformers --slow
+        run: python -m pytest --pyargs curated_transformers
```
danieldk, line 88: Removing the cache sounds great, but maybe we want to keep the slow markers as a middle ground for the future? Or maybe rename the existing `veryslow` tests to `slow`.

shadeMe: Can't argue with that. I'll rename `veryslow` to `slow` and update the BuildKite pipeline.

danieldk (approved):

PR #210, opened 2023-06-29
Add support for the `tokenizers` package

danieldk:
## Description

This change adds support for the `tokenizers` package. We started implementing support for reading `tokenizer.json` files. However, fully supporting this format is currently not our focus, so instead we use the `tokenizers` package to support these tokenizers.

To support this, the class hierarchy needs to be restructed a bit:

* The original `Tokenizer` class is split into two classes: `TokenizerBase` defines the interface for tokenizers. `LegacyTokenizer` implements the interface for our original tokenizers.
* The new `Tokenizer` class it the `tokenizers`-based tokenizer. This tokenizer has a prominent name, because going forward it will be the preferred/default tokenizer.

### Types of change

Feature

## Checklist


curated_transformers/tokenization/tokenizer.py
```diff
@@ -73,6 +83,172 @@ def padded_tensor(self, *, padding_id: int, pad_left: bool = False):
...
+            seq_pieces = []
+
+            for chunk in seq:
+                if isinstance(chunk, MergedSpecialPieceChunk):
+                    piece_id = self.tokenizer.token_to_id(chunk.piece)
+                    seq_ids.append(piece_id)
+                    seq_pieces.append(chunk.piece)
+                else:
+                    encoding = self.tokenizer.encode(
+                        chunk.text,  # add_special_tokens=False
```
shadeMe, line 204: Intentionally commented out?

setup.cfg
```diff
@@ -16,6 +16,7 @@ python_requires = >=3.8
 install_requires =
     curated-tokenizers>=0.0.7,<0.1.0
     huggingface-hub>=0.14
+    tokenizers
```
shadeMe, line 19: No pins on this requirement?

curated_transformers/tokenization/tokenizer.py
```diff
@@ -73,6 +83,172 @@ def padded_tensor(self, *, padding_id: int, pad_left: bool = False):
...
+
+        decoded = [clean_up_decoded_string_like_hf(string) for string in decoded]
+
+        return decoded
+
+    def encode(
+        self, input: Union[Iterable[InputChunks], Iterable[str]]
+    ) -> PiecesWithIds:
+        input = cast(Union[List[InputChunks], List[str]], list(input))
+        if isinstance(input[0], str):
```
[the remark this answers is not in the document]
danieldk: I find these decision difficult, one could argue that this is the role of type checking. But in this case the errors are really ugly if you use incorrect input, so I added validation.

curated_transformers/tokenization/tokenizer.py
danieldk: I think that was probably during debugging, removed comment.

setup.cfg
danieldk: Set the lower version to the current version.

shadeMe (approved):

PR #213, opened 2023-07-05
Fallback to the default device during `bitsandbytes` layer init

shadeMe:
## Description

Currently, we error out if the user attempts to use `bitsandbytes` quantization without having [our fork](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of the library installed. This PR removes that restriction at the expense of some extra memory usage.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,5 +46,6 @@ pip install curated-transformers
 to build the library from source by cloning the [following branch](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of our fork. Installation instructions
 can be found [here](https://github.com/shadeMe/bitsandbytes/blob/linear-layer-device/compile_from_source.md).
 
-Users can still use the quantization feature without building the `bitsandbytes` library from our fork. However, they may encounter increased memory usage during the model
-initialization phase. The extra memory will be freed once the model has been successfully quantized.
+Users can still use the quantization feature without building the `bitsandbytes` library from our fork. However, this will result in increased CPU memory usage during the model
+initialization phase. The extra overhead amounts to about 2x the memory required for each model parameter, but this value is not cumulative as parameters are deserialized one-by-one.
+GPU memory usage remains unaffected.
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ pip install curated-transformers
 
 ### Installation
 
-Since `curated-transformers` requires functionality that isn't currently present in `bitsandbytes` (as of `v0.39.0`), one needs to install the library
-from source by cloning the [following branch](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of our fork. Installation instructions
+`curated-transformers` requires functionality that isn't currently present in `bitsandbytes` (as of `v0.39.1`). While the said functionality is optional, we recommend users
+to build the library from source by cloning the [following branch](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of our fork. Installation instructions
 can be found [here](https://github.com/shadeMe/bitsandbytes/blob/linear-layer-device/compile_from_source.md).
+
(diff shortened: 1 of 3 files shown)


curated_transformers/quantization/bnb/impl.py
```diff
@@ -166,8 +173,10 @@ def _assert_bitsandbytes_installed():
...
         )
     elif not has_bitsandbytes_linear_device:
-        raise ValueError(
-            "The currently installed `bitsandbytes` library does not support "
-            "passing device parameters to its modules. Refer to the readme "
-            "for more information on the requirements"
+        msg = (
+            "The installed version of the `bitsandbytes` library does not support passing "
+            "device parameters to its modules. This can result in increased memory usage "
+            "during model initialization. Please refer to the readme for more information."
```
danieldk, line 179:
```suggestion
            "during model initialization. Please refer to the README for more information."
```

README.md
```diff
@@ -42,6 +42,9 @@ pip install curated-transformers
 
 ### Installation
 
-Since `curated-transformers` requires functionality that isn't currently present in `bitsandbytes` (as of `v0.39.0`), one needs to install the library
-from source by cloning the [following branch](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of our fork. Installation instructions
+`curated-transformers` requires functionality that isn't currently present in `bitsandbytes` (as of `v0.39.1`). While the said functionality is optional, we recommend users
+to build the library from source by cloning the [following branch](https://github.com/shadeMe/bitsandbytes/tree/linear-layer-device) of our fork. Installation instructions
 can be found [here](https://github.com/shadeMe/bitsandbytes/blob/linear-layer-device/compile_from_source.md).
+
+Users can still use the quantization feature without building the `bitsandbytes` library from our fork. However, they may encounter increased memory usage during the model
```
danieldk, line 49: This is a vague. I am trying to remember what happens without the `device` argument. Without `device`, we cannot initially place the layers on the `meta` device, meaning that the layers will get initialized first (in 4/8-bit). So AFAIR, there are two issues:

* Unnecessary initialization is done. This increases model load/initialization time quite a bit, but is probably not really an issue wrt. inference time.
* Memory use: suppose that we have an BFloat16 key matrix. With the meta device we have at most bfloat16_size(key) + quantized_size(key) in memory (the matrix gets quantized, after quantization, we emplace the meta device tensor). Without the meta device we have at most bfloat16_size(key) + 2 * quantized_size(key). The latter because before emplacement, there is the quantized key matrix and the randomly initialized matrix. However, since we are only quantizing one set of parameters at the time, the extra overhead is basically quantized_size(largest_parameter).

Is that analysis of the memory use correct? I think the largest quantizable parameter is usually the embedding matrix?

shadeMe: > Unnecessary initialization is done. This increases model load/initialization time quite a bit, but is probably not really an issue wrt. inference time.

Correct. 

As for memory usage, there is no distinction between `quantized_size` and `bfloat16_size` during model initialization: The `bitsandbytes` parameters are internally stored as half-precision floats on the CPU and only get quantized to `int8`/`fp4` when they are moved to the CUDA device.

For every parameter, we have:
* With the `meta` device:  The deserialized `bfloat16` tensor on the CPU + the meta `float16` tensor in the quantized layer (zero-sized).
* Without the `meta` device: The deserialized `bfloat16` tensor on the CPU + the `float16`[1] tensor in the quantized layer + extra overhead added by the initialization code.

At the time of emplacement, we move the deserialized `bfloat16` tensor into the quantized layer and then upload it to the GPU. This should release the host memory that was originally occupied by the former. Ultimately, only the quantized weights make it to the GPU.

[1] - We don't set the default `torch` tensor type to the model's `dtype` before constructing the model, so we're probably also incurring the cost of performing the initialization in full precision and then downcasting it. This is something we can/should fix, though.

danieldk (approved):

PR #215, opened 2023-07-06
Add a default generator

danieldk:
## Description

`DefaultGenerator` can be used when a generation model does not require custom prompting or other model-specific settings such as a non-standard end-of-sequence identifier.

This change reimplements `DollyV2Generator` and `FalconGenerator` using `DefaultGenerator`.

The `AutoGenerator` class it moved to the `generation` package to avoid import cycles.

### Types of change

Feature

## Checklist

### DIFF
--- a/curated_transformers/generation/auto_generator.py
+++ b/curated_transformers/generation/auto_generator.py
@@ -1,0 +1,70 @@
+from typing import Dict, Optional, Type, cast
+
+import torch
+
+from curated_transformers.generation.default_generator import DefaultGenerator
+from curated_transformers.generation.hf_hub import FromHFHub
+
+from ..quantization import BitsAndBytesConfig
+from ..util.auto_model import AutoModel
+from .dolly_v2 import DollyV2Generator
+from .falcon import FalconGenerator
+from .generator_wrapper import GeneratorWrapper
+
+# For the time being, we enable support for a generator on a case-by-case basis.
+# In the future we might defer all unknown generators to DefaultGenerator.
+GENERATOR_MAP: Dict[str, Type[FromHFHub]] = {
+    "dolly-v2": DollyV2Generator,
+    "falcon": FalconGenerator,
+    "llama": DefaultGenerator,
+}
+
+
+class AutoGenerator(AutoModel[GeneratorWrapper]):
+    """Causal LM generator loaded from the Hugging Face Model Hub.
+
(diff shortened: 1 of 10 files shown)


curated_transformers/generation/auto_generator.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class AutoGenerator(AutoModel[GeneratorWrapper]):
+    """Causal LM generator loaded from the Hugging Face Model Hub.
+
+    **NOTE** - This class can currently only be used with the following models:
+        - Models based on Dolly v2 (contain ``dolly-v2``` in the name).
+        - Models based on Falcon (contain ``falcon`` in the name).
+        - Models based on LLaMA (contain ``llama`` in the name).
+    """
```
shadeMe, line 30:
```suggestion
    """
     .. highlights::
    	Causal LM generator loaded from the Hugging Face Model Hub.

        .. attention::
          	This class can currently only be used with the following models:
          	 - Models based on Dolly v2 (contain ``dolly-v2`` in the name).
          	 - Models based on Falcon (contain ``falcon`` in the name).
          	 - Models based on LLaMA (contain ``llama`` in the name).
    """
```
To keep in line with the upcoming docs PR.

curated_transformers/generation/default_generator.py
```diff
@@ -0,0 +1,100 @@
...
+from .string_generator import StringGenerator
+
+# Only provided as typing.Self in Python 3.11+.
+Self = TypeVar("Self", bound="DefaultGenerator")
+
+
+class DefaultGenerator(GeneratorWrapper, FromHFHub):
+    """
+    Generator wrapper for models that do not need specific prompting.
+    """
```
shadeMe, line 25:
```suggestion
    """
    .. highlights::
    	Generator wrapper for models that do not need specific prompting.
    """
```

curated_transformers/generation/auto_generator.py
danieldk: I am not sure I understand these? `highlights` is for block quotes?

https://docutils.sourceforge.io/docs/ref/rst/directives.html#highlights

shadeMe: As discussed elsewhere, we can skip the `.. highlights::` annotation in favour of a global Sphix config setting.

danieldk: Added the attention block.

PR #224, opened 2023-07-11
Add an example that rewrites sentences in the passive voice to active voice

danieldk:
## Description

This adds another example besides the more generic QA-like generation script that we already have.

We use `FalconGenerator` rather than `AutoGenerator` because the prompt was created with Falcon 7B with instruction tuning.

### Types of change

Example.

## Checklist

### DIFF
--- a/examples/passive2active.py
+++ b/examples/passive2active.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+
+# Rewrite sentences in the passive voice to the active voice using the
+# 7B parameter Falcon model. This is only a small demo of using Curated
+# Transformers for a specific application of generation. The example only
+# works for relatively simple sentences, use larger models/better prompts/
+# finetuning to get better results.
+#
+# Example usage:
+#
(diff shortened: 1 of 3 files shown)


examples/passive2active.py
```diff
@@ -0,0 +1,145 @@
...
+
+[{"passive": "The medal was won by the Dutch speed skater.", "active": "..."},
+ {"passive": "Anita was driven to the theatre by Carla.", "active": "..."}]'"""
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    description="Rewrite sentences in the active voice to passive voice",
+    epilog=EPILOG,
+)
+parser.add_argument(
```
shadeMe, line 43: Shall we also add an argument to enable quantization?

danieldk: Great idea. Added. Also had to make `RefinedWebModel` inherit `Quantizable`.

shadeMe (approved):

PR #245, opened 2023-07-14
Add `VocabMaskTransform`

shadeMe:
## Description

Add `VocabMaskTransform` logit transform that's unconditionally applied to all generations.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/curated_transformers/generation/config.py
+++ b/curated_transformers/generation/config.py
@@ -21,21 +21,24 @@ class GeneratorConfig(ABC):
     """
     Configuration of the generator.
 
+    :param default_logits_transform:
+        The logits transform that are unconditionally applied to all
+        generations. Usually multiple composed transforms, can be
+        empty.
     :param eos_id:
         End-of-sequence identifier that should end the generation of a sequence
         when predicted. When this value is set to `None`, it is the
         responsibility of the generator to set it.
-
     :param max_generated_pieces:
         The maximum number of generation steps. This condition is a noop
         for values less than 1. When this value is set to `None`, it is the
         responsibility of the generator to set it.
     """
 
+    default_logits_transform: LogitsTransform = CompoundLogitTransforms([])
     eos_id: Optional[int] = None
     max_generated_pieces: Optional[int] = None
 
-    @abstractmethod
     def logits_transform(self) -> LogitsTransform:
         """
         Get logit transform for the configuration.
@@ -43,7 +46,7 @@ class GeneratorConfig(ABC):
         :returns:
             Logits transform. Usually multiple composed transforms.
         """
-        ...
+        return self.default_logits_transform
 
     def stop_condition(self) -> StopCondition:
         """
@@ -70,6 +73,7 @@ class GeneratorConfig(ABC):
         return conditions
 
 
+@dataclass
 class GreedyGeneratorConfig(GeneratorConfig):
     """
     Configuration for greedy generation.
@@ -78,8 +82,7 @@ class GreedyGeneratorConfig(GeneratorConfig):
     to deterministic generation.
     """
 
-    def logits_transform(self) -> LogitsTransform:
-        return CompoundLogitTransforms([])
+    pass
 
 
 @dataclass
@@ -108,6 +111,7 @@ class SampleGeneratorConfig(GeneratorConfig):
     def logits_transform(self) -> LogitsTransform:
         return CompoundLogitTransforms(
             [
+                self.default_logits_transform,
                 TemperatureTransform(self.temperature),
                 TopKTransform(self.top_k),
             ]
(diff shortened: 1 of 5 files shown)


curated_transformers/generation/config.py
```diff
@@ -21,29 +21,34 @@ class GeneratorConfig(ABC):
...
         when predicted. When this value is set to `None`, it is the
         responsibility of the generator to set it.
-
     :param max_generated_pieces:
         The maximum number of generation steps. This condition is a noop
         for values less than 1. When this value is set to `None`, it is the
         responsibility of the generator to set it.
     """
 
+    default_logits_transform: LogitsTransform = field(
```
danieldk, line 38: This is not how configuration has worked so far. In the current setup, we'd rather add a field like `masked_pieces: Set[int]` and the `logits_transform` method of the particular config would convert this to a `LogitsTransform`.

We should make it possible for the user to directly configure transforms (I have an idea of how we could do this, but this requires changing the class hierarchy, but that's out the scope of this PR). But the main idea behind the current configuration classes is that the user configures concrete options and does not have to bother with the underlying classes (like `LogitsTransform` subclasses).

curated_transformers/generation/logits.py
```diff
@@ -121,3 +122,43 @@ def _process_logits(self, logits: Tensor):
             return
 
         logits /= self.temperature
+
+
+class MaskTransform(LogitsTransform):
```
danieldk, line 127: I think it would be nice if the name also makes clear what we are masking, since we already have multiple masks. Maybe `MaskPiecesTransform`? `VocabMaskTransform`?

curated_transformers/generation/logits.py
```diff
@@ -121,3 +122,43 @@ def _process_logits(self, logits: Tensor):
...
+    """
+    Set the probability of specific classes to zero.
```
danieldk, line 129:
```suggestion
    Set the probability of specific vocabulary items to zero.
```

curated_transformers/generation/logits.py
```diff
@@ -121,3 +122,43 @@ def _process_logits(self, logits: Tensor):
...
+    """
+
+    def __init__(self, classes_to_mask: List[int]):
```
danieldk, line 132:
```suggestion
    def __init__(self, classes_to_mask: List[int]):
```

Probably `Set[int]` or if we want to be more generic `Iterable[int]`.

curated_transformers/generation/logits.py
```diff
@@ -121,3 +122,43 @@ def _process_logits(self, logits: Tensor):
...
+        :param classes_to_mask:
+            Identifers pertaining to the classes that need to be masked.
+            An empty list results in a no-op.
+        """
+        super().__init__()
+        self.classes_to_mask = torch.tensor(
+            classes_to_mask, dtype=torch.long
+        )  # `torch.int32` isn't supported in indexing operations prior in torch<2.0.0.
+        if self.classes_to_mask.dim() != 1:
+            raise ValueError("Class selector tensor for mask transform must be 1D")
```
danieldk, line 145: I think in the case of generation the classes are always vocab items, so I think it's clearer if the errors mention that. Something like:

```suggestion
            raise ValueError("Vocabulary items mask must be 1D")
```

the same of all the mentions of _class_ below.

curated_transformers/generation/logits.py
```diff
@@ -121,3 +122,43 @@ def _process_logits(self, logits: Tensor):
...
+    def _process_logits(self, logits: Tensor):
+        if self.classes_to_mask.size(dim=-1) == 0:
+            return
+
+        max_expected = logits.size(dim=-1)
+        max_received = int(self.classes_to_mask.max(dim=-1).values)
+        if max_received >= max_expected:
+            raise ValueError(
+                f"Class identifiers for mask transform must be < {max_expected}, but got {max_received}"
+            )
```
danieldk, line 161: I think Torch reports out of bound indexing. Good error messages are nice, but this does launch another kernel. I think it's worth looking wether the indexing below can be wrapped in a `try...except`, so that we don't have to go over the array twice.

curated_transformers/generation/config.py
shadeMe: Makes sense. I'll rewrite it as suggested.

curated_transformers/generation/logits.py
shadeMe: Renamed to `VocabMaskTransform`.

curated_transformers/generation/logits.py
shadeMe: Switched to `Iterable`

curated_transformers/generation/logits.py
shadeMe: Fixed.

curated_transformers/generation/logits.py
shadeMe: Switched to `try...except` .

danieldk (approved):

PR #248, opened 2023-07-17
Replace decoder layers by a shared layer

danieldk:
## Description

This change replaces the decoder layers by a shared decoder layer, `DecoderLayer`. Since the overlap between encoder and decoder layers is very large, both are factored out to a single `_TransformerLayer`.

The `_TransformerLayer` class is not public because exposing its `forward` method would violate the Liskov substitution principle, but it holds the implementation of both types of layers.

The Falcon decoder layer does not currently use the shared decoder layer because its dropout use is nonstandard. However, I might try to support this as well in a follow-up PR.

This change also updates the Dolly V2 tests. The different order of summing in the residual connection leads to different outputs due to the imprecision of `bfloat16`.

### Types of change

Feature: improve curated as a building block.

## Checklist


curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+        parallel_attention: bool,
+        qkv_head_sharing: QkvHeadSharing,
+        qkv_mode: QkvMode,
+        rotary_embeds: Optional[RotaryEmbeddingConfig],
+        use_bias: bool,
+        use_gate: bool,
+        device: Optional[torch.device] = None
+    ):
+        """
+        Construct an decoder layer.
```
shadeMe, line 238:
```suggestion
        Construct a decoder layer.
```

curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+    def forward(
+        self,
+        input: Tensor,
+        attention_mask: Optional[AttentionMask],
+        cache: Optional[KeyValueCache] = None,
+        positions: Optional[Tensor] = None,
+        store_cache: bool = False,
+    ) -> Tuple[Tensor, Optional[KeyValueCache]]:
+        """
+        Apply the transformer layer to the given piece hidden representations.
```
shadeMe, line 300:
```suggestion
        Apply the decoder layer to the given piece hidden representations.
```

curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+        parallel_attention: bool,
+        qkv_head_sharing: QkvHeadSharing,
+        qkv_mode: QkvMode,
+        rotary_embeds: Optional[RotaryEmbeddingConfig],
+        use_bias: bool,
+        use_gate: bool,
+        device: Optional[torch.device] = None
+    ):
+        """
+        Construct an decoder layer.
```
shadeMe, line 355:
```suggestion
        Construct an encoder layer.
```

curated_transformers/models/albert/layer_group.py
```diff
@@ -26,6 +27,12 @@ def __init__(
     ) -> None:
         super().__init__()
 
+        layer_norm = partial(
```
shadeMe, line 30: Was this extracted into a `partial` just to avoid duplicate code in the `EncoderLayer` callsite?

curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+        qkv_head_sharing: QkvHeadSharing,
+        qkv_mode: QkvMode,
+        rotary_embeds: Optional[RotaryEmbeddingConfig],
+        use_bias: bool,
+        use_causal_mask: bool,
+        use_gate: bool,
+        device: Optional[torch.device] = None
+    ):
+        """
+        Construct an decoder layer.
```
shadeMe, line 79:
```suggestion
        Construct a transformer layer.
```

curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+)
+from .feedforward import PointwiseFeedForward
+
+
+@dataclass
+class TransformerLayerNorms:
+    """
+    Layer normalizations used in a transformer layer.
+
+    By default all the normalizations are disabled by setting the layer
```
shadeMe, line 24:
```suggestion
    By default, all the normalizations are disabled by setting the layer
```

curated_transformers/layers/transformer.py
```diff
@@ -0,0 +1,431 @@
...
+    normalization to the Torch `Identity` module. Therefore, only
```
shadeMe, line 25:
```suggestion
    normalization to the Torch ``Identity`` module. Therefore, only
```

curated_transformers/models/albert/layer_group.py
danieldk: Yeah, shortens it a considerable bit. The lines are already pretty long, so the arguments get spread over multiple lines.

shadeMe (approved):

PR #252, opened 2023-07-18
Add support for ALiBi in self-attention layer

shadeMe:
## Description

ALiBi support in Falcon/RefinedWebModel will be enabled in a follow-up PR once their related refactoring is complete.

### Types of change

Enhancement

## Checklist


curated_transformers/layers/attention.py
```diff
@@ -173,23 +174,140 @@ class QkvMode(IntEnum):
...
+        # to avoid previously picked values.
+        if closest_pow_2 < num_attention_heads:
+            new_ratio_base = 2.0 ** (-4.0 / closest_pow_2)
+            num_remaining_heads = min(
+                closest_pow_2, num_attention_heads - closest_pow_2
+            )
+            new_ratios = torch.pow(
+                new_ratio_base, torch.arange(1, 1 + 2 * num_remaining_heads, 2)
+            )
+            ratios = torch.cat([ratios, new_ratios])
```
danieldk, line 230: Could you explain why this is needed? From the paper, it seems that the slopes can be computed as:

```suggestion
        ratio = 2 ** (-8/num_attention_heads)
        ratios = ratio ** torch.arange(1, num_attention_heads + 1)
```

curated_transformers/layers/attention.py
```diff
@@ -407,6 +546,11 @@ def forward(
             )
 
         if _TORCH_SDP.get():
+            if self.use_alibi:
+                msg = "ALiBi disabled when using PyTorch SDP"
+                warnings.filterwarnings(action="once", message=msg)
+                warnings.warn(msg)
+
```
danieldk, line 553: I think this should be an exception, since it would return incorrect outputs. However, I think we could support ALiBi with Torch SDP. The `attn_mask` argument of `F.scaled_dot_product_attention` is additive when it's a float tensor. ALiBi is also additive, so we could make it work by adding the `ALiBi` biases to the logit mask for positions that are not masked out.

curated_transformers/layers/attention.py
```diff
@@ -173,23 +174,153 @@ class QkvMode(IntEnum):
...
+            Maximum number of timesteps to calculate.
+        :returns:
+            Multi-headed linear bias tensor.
+
+            *Shape:* ``(1, heads, seq_len, seq_len)`` (non-causal) or
+            ``(1, heads, 1, seq_len)`` (causal)
+
+        :meta private:
+        """
+        if not self.is_causal:
```
danieldk, line 265: Can we flip the branches around, then we can have `if self.is_causal:`

shadeMe: Done.

curated_transformers/layers/attention.py
shadeMe: Added.

curated_transformers/layers/attention.py
```diff
@@ -407,15 +542,27 @@ def forward(
...
+            attn_mask = (
+                None if combined_mask is None else combined_mask.logit_mask(query.dtype)
+            )
+
+            # Add AliBi to the logit mask
+            if self.use_alibi and attn_mask is not None:
+                assert combined_mask is not None
+                assert self.attention.linear_biases is not None
+                bool_mask = combined_mask.bool_mask
+                biases = self.attention.linear_biases.calculate_biases(key.size(-2)).to(
```
danieldk, line 554:
```suggestion
                biases = self.attention.linear_biases.calculate_biases(query.size(-2)).to(
```

Since the number of key heads can be 1.

curated_transformers/layers/attention.py
```diff
@@ -407,15 +542,27 @@ def forward(
             )
 
         if _TORCH_SDP.get():
+            attn_mask = (
+                None if combined_mask is None else combined_mask.logit_mask(query.dtype)
+            )
+
+            # Add AliBi to the logit mask
+            if self.use_alibi and attn_mask is not None:
```
danieldk, line 550: What if `attn_mask is None` and we use ALiBi?

curated_transformers/layers/attention.py
shadeMe: We're using the key length here, so the number of heads shouldn't matter.

curated_transformers/layers/attention.py
```diff
@@ -1,4 +1,5 @@
 import math
+import warnings
```
danieldk, line 2:
```suggestion
```

Not needed anymore.

PR #257, opened 2023-07-21
Make some changes to support `torch.compile` and add tests

danieldk:
## Description

Supporting `torch.compile` requires some small changes:

- We can't use `IntEnum`, replace single use by `Enum`.
- In the causal mask generation, the data type gets inferred incorrectly, explicitly give the `dtype`.

This change also adds tests for `torch.compile`. To avoid repetitiveness, factor out decoder tests similar to encoder tests.

In the `torch.compile` test for NeoX we disable testing with cache and positions. `torch.compile` is tripping over the rotary embeddings code where the rotary embeddings are only applied to part of the query/key representations.

While debugging, I also found that `GeluNew` can be implemented with fewer operations.

### Types of change

Feature/tests

## Checklist

### DIFF
--- a/curated_transformers/layers/activations.py
+++ b/curated_transformers/layers/activations.py
@@ -51,11 +51,7 @@ class GeluFast(Module):
 
             *Shape:* ``(batch_size, seq_len, width)``
         """
-        return (
-            0.5
-            * input
-            * (
-                1.0
-                + torch.tanh(input * 0.7978845608 * (1.0 + 0.044715 * input * input))
-            )
-        )
+        alpha = math.sqrt(2.0 / math.pi)
+        beta = 0.044715
+
+        return 0.5 * input * (1.0 + torch.tanh(alpha * (input + beta * input.pow(3))))
--- a/curated_transformers/tests/models/gpt_neox/test_decoder.py
+++ b/curated_transformers/tests/models/gpt_neox/test_decoder.py
@@ -1,103 +1,36 @@
 import pytest
-import torch
 
-from curated_transformers.layers.attention import AttentionMask
 from curated_transformers.models.gpt_neox.decoder import GPTNeoXDecoder
-from curated_transformers.tests.util import torch_assertclose
 
-from ...compat import has_hf_transformers, transformers
+from ...compat import has_hf_transformers, has_torch_compile
 from ...conftest import TORCH_DEVICES
+from ..util import assert_decoder_output_equals_hf
 
 
 @pytest.mark.skipif(not has_hf_transformers, reason="requires huggingface transformers")
 @pytest.mark.parametrize("torch_device", TORCH_DEVICES)
 def test_decoder(torch_device):
-    hf_model = transformers.AutoModel.from_pretrained(
-        "trl-internal-testing/tiny-random-GPTNeoXForCausalLM"
+    assert_decoder_output_equals_hf(
+        GPTNeoXDecoder,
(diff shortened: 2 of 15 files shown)


curated_transformers/layers/activations.py
```diff
@@ -51,11 +51,7 @@ def forward(self, input: Tensor) -> Tensor:
...
-            * input
-            * (
-                1.0
-                + torch.tanh(input * 0.7978845608 * (1.0 + 0.044715 * input * input))
-            )
-        )
+        alpha = math.sqrt(2.0 / math.pi)
+        beta = 0.044715
+
+        return 0.5 * input * (1.0 + torch.tanh(alpha * (input + beta * input.pow(3))))
```
shadeMe, line 57: Is there a reference for this?

curated_transformers/tests/models/gpt_neox/test_decoder.py
```diff
@@ -1,103 +1,36 @@
...
-    torch.manual_seed(0)
-    X = torch.randint(0, hf_model.config.vocab_size, (2, 10), device=torch_device)
-    # NOTE: this test may break in the future because we are relying on a
-    #       slicing bug, which results in more general support for positions
-    #       (like our implementation):
-    #
-    #       https://github.com/huggingface/transformers/blob/f924df3c7e5150317ef47754a31cebf2893570ce/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L278
+def test_decoder_with_torch_compile(torch_device):
+    # NOTE: testing with cache/positions fails. All outputs seems ok until
+    # the concatenation of the vectors with/without rotary embeddings:
```
shadeMe, line 26: Shall we (temporarily) add an `xfail` test case for that?

curated_transformers/tests/models/util.py
```diff
@@ -9,20 +9,152 @@
...
+
+    torch.manual_seed(0)
+    X = torch.randint(0, hf_model.config.vocab_size, (2, 10), device=torch_device)
+
+    with torch.no_grad():
+        Y = model(X).logits
+        Y_hf = hf_model(X).logits
+    torch_assertclose(Y, Y_hf, atol=atol, rtol=rtol)
+
+    mask = torch.rand((2, 10), dtype=torch.float, device=torch_device) < 0.5
```
shadeMe, line 48: Do we need to reset the manual seed for this invocation as well? I recall that we do that in the sampling generator tests.

curated_transformers/layers/activations.py
danieldk: No, I just rewrote the original equation a bit (multiplying out stuff).

curated_transformers/tests/models/util.py
danieldk: I considered doing this, since it's a known failure. However, the problem is that if Torch fixes this issue, the test will suddenly pass (and thus fail), which is probably not nice.

curated_transformers/tests/models/gpt_neox/test_decoder.py
danieldk: I considered doing this, since it's a known failure. However, the problem is that if Torch fixes this issue, the test will suddenly pass (and thus fail), which is probably not nice.

curated_transformers/tests/models/util.py
danieldk: That is so that we can reorder/remove particular generation tests in the future without failing the generation tests after that. But that's not an issue here, since we do not manually validate the output.

PR #263, opened 2023-07-26
Add support for LLaMA2

danieldk:
## Description

LLaMA2 support different number of query and key-value heads. Also add new models to test k-n and k-1 head configurations.

### Types of change

Feature

## Checklist


shadeMe (approved):
Tests pass locally, but we need to increment the `transformers` pin in the CI.

EDIT: We'll also need to add a separate [generator](https://github.com/facebookresearch/llama/blob/main/llama/generation.py) for the chat models as they need custom prompting.

PR #266, opened 2023-07-27
Make `DataclassAsTuple` types actual tuples

danieldk:
## Description

Before this change, `DataclassAsTuple` provided an `astuple` method to convert the dataclass to a tuple. This change makes `DataClassAsTuple` a subclass of `tuple`, so that it can be used directly in TorchScript tracing.

This removes the need to conditionally return tuples depending on whether we are tracing.

We also make `KeyValueCache` derive from `DataclassAsTuple` rather than `DataclassAsDict`. This provides weaker guarantees in rewrapping (since) we don't get keys anymore. However, the benefit is that we can run tracing with strictness switched on (since the return value does not contain a dict).

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/curated_transformers/layers/cache.py
+++ b/curated_transformers/layers/cache.py
@@ -78,7 +78,7 @@ class KeyValueCache(DataclassAsTuple):
             or not all(isinstance(item, Tensor) for item in key_value_cache)
         ):
             raise ValueError(
-                f"Key-value cache is not of the `KeyValueCache` type, nor Tuple[Tensor, Tensor]: {type(key_value_cache).__name__}"
+                f"Key-value cache is not of the `KeyValueCache` type, nor `Tuple[Tensor, Tensor]`: `{type(key_value_cache).__name__}`"
             )
 
         key_cache = key_value_cache[0]
--- a/curated_transformers/layers/cache.py
+++ b/curated_transformers/layers/cache.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Dict, Optional, Protocol, Type, TypeVar, Union
+from typing import Dict, Optional, Protocol, Tuple, Type, TypeVar, Union
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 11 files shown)


curated_transformers/util/dataclass.py
```diff
@@ -55,47 +55,90 @@ def __setitem__(self, key: str, value: Tensor) -> None:
...
+        elif isinstance(value, list):
+            if all(isinstance(item, Tensor) for item in value):
+                return tuple(value)
+            elif all(isinstance(item, tuple) for item in value):
+                return tuple(value)
             else:
+                type_names = ", ".join(sorted({type(item).__name__ for item in value}))
                 raise TypeError(
-                    f"Field '{field.name}' has unsupported type `{type(value).__name__}`"
+                    f"List must be List[Tensor] or List[Tuple[...]], found types: {type_names}"
```
shadeMe, line 131:
```suggestion
                    f"List must be `List[Tensor]` or `List[Tuple[...]]`, found types: {type_names}"
```

curated_transformers/layers/cache.py
```diff
@@ -72,16 +72,21 @@ def jit_rewrap(
...
-        key = key_value_cache.get("key")
-        if key is None:
+        if (
+            not isinstance(key_value_cache, tuple)
+            or len(key_value_cache) != 2
+            or not all(isinstance(item, Tensor) for item in key_value_cache)
+        ):
             raise ValueError(
-                "Key-value cache is not of the `KeyValueCache` type, nor a dict with 'key'."
+                f"Key-value cache is not of the `KeyValueCache` type, nor Tuple[Tensor, Tensor]: {type(key_value_cache).__name__}"
```
shadeMe, line 81:
```suggestion
                f"Key-value cache is not of the `KeyValueCache` type, nor `Tuple[Tensor, Tensor]`: `{type(key_value_cache).__name__}`"
```

shadeMe (approved):
LGTM! Minor fixes.

PR #270, opened 2023-07-27
Documentation fixes

shadeMe:
### Types of change

Docs

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,14 @@ Supported encoder-only models:
 Supported decoder-only models:
 
 - GPT-NeoX
-- LLaMA
+- LLaMA 1/2
 - Falcon
 
 Generator wrappers:
 
 - Dolly v2
 - Falcon
+- LLaMA 1/2
 
 All types of models can be loaded from Huggingface Hub.
 
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ spaCy integration for curated transformers is provided by the
 [`spacy-curated-transformers`](https://github.com/explosion/spacy-curated-transformers)
 package.
 
+## ⚠️ Warning: Tech Preview
+
+Curated Transformers 0.9.x is a tech preview, we will release Curated Transformers
+1.0.0 with a stable API and semver guarantees over the coming weeks.
+
 ## ⏳ Install
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -59,11 +59,6 @@ spaCy integration for curated transformers is provided by the
 [`spacy-curated-transformers`](https://github.com/explosion/spacy-curated-transformers)
 package.
 
-## ⚠️ Warning: Tech Preview
-
-Curated Transformers 0.9.x is a tech preview, we will release Curated Transformers
-1.0.0 with a stable API and semver guarantees over the coming weeks.
-
 ## ⏳ Install
 
 ```bash
--- a/curated_transformers/generation/falcon.py
+++ b/curated_transformers/generation/falcon.py
@@ -1,6 +1,8 @@
 from typing import List, TypeVar
 
+from ..models.falcon import FalconCausalLM
 from ..tokenizers.chunks import InputChunks, TextChunk
+from ..tokenizers.tokenizer import Tokenizer
 from .default_generator import DefaultGenerator
 from .hf_hub import FromHFHub
 
@@ -12,6 +14,18 @@ class FalconGenerator(DefaultGenerator, FromHFHub):
     """
     Generator for Falcon model variants.
     """
+
+    def __init__(self, tokenizer: Tokenizer, causal_lm: FalconCausalLM):
+        """
+        Construct a Falcon generator.
+
+        :param tokenizer:
+            A Falcon tokenizer.
+        :param causal_lm:
+            A Falcon causal language model.
+
+        """
+        super().__init__(tokenizer, causal_lm)
 
     def preprocess_prompts(self, prompts: List[str]) -> List[InputChunks]:
         return [
--- a/curated_transformers/layers/attention.py
+++ b/curated_transformers/layers/attention.py
@@ -50,17 +50,14 @@ class AttentionMask(DataclassAsDict):
     Mask for attention calculation. Sequence elements for which the
     corresponding mask element is set to ``False`` are ignored during
     attention calculation.
+
+    :param bool_mask:
+        The boolean mask.
     """
 
     bool_mask: Tensor
 
     def __init__(self, bool_mask: Tensor):
-        """
-        Construct an attention mask.
-
-        :param bool_mask:
-            The boolean mask.
-        """
         if bool_mask.dtype != torch.bool:
             raise ValueError("Expected the attention mask to be of dtype 'torch.bool'")
 
@@ -84,7 +81,7 @@ class AttentionMask(DataclassAsDict):
     ) -> Optional["AttentionMask"]:
         """
         Rewrap TorchScript dictionary conversion of an attention mask
-        as an `AttentionMask`.
+        as an ``AttentionMask``.
 
         :param attention_mask:
             The attention mask or its dictionary representation. If the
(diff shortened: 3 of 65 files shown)


README.md
```diff
@@ -59,11 +59,6 @@ spaCy integration for curated transformers is provided by the
 [`spacy-curated-transformers`](https://github.com/explosion/spacy-curated-transformers)
 package.
 
-## ⚠️ Warning: Tech Preview
```
danieldk, line 62: We should keep this until the release next week. This is the landing page that people will see, even if this is the `main` branch.

curated_transformers/generation/falcon.py
```diff
@@ -13,6 +15,18 @@ class FalconGenerator(DefaultGenerator, FromHFHub):
     Generator for Falcon model variants.
     """
 
+    def __init__(self, tokenizer: Tokenizer, causal_lm: FalconCausalLM):
```
danieldk, line 18: This is nice for the documentation, but I think we should keep the classes minimal, so no constructor if it's effectively the same as the superclass constructor. If the generated documentation is not clear, we should change the documentation of the superclass constructor.

I don't think people will use the constructor in practice, but instead load it up with a static factory method.

curated_transformers/layers/attention.py
```diff
@@ -148,12 +145,34 @@ def filter_batch_items(self, mask: Tensor) -> "AttentionMask":
...
 
     def dim(self) -> int:
+        """
+        Return the number of dimensions in the mask.
+        """
         return self.bool_mask.dim()
 
     def merge_mask(self, other: "AttentionMask") -> "AttentionMask":
+        """
+        Merge another attention mask with this one.
```
danieldk, line 155:
```suggestion
        Merge this attention mask with another attention mask.
```

curated_transformers/layers/attention.py
```diff
@@ -148,12 +145,34 @@ def filter_batch_items(self, mask: Tensor) -> "AttentionMask":
...
+        """
         return AttentionMask(self.bool_mask.logical_and(other.bool_mask))
 
-    def logit_mask(self, dtype: torch.dtype):
+    def logit_mask(self, dtype: torch.dtype) -> Tensor:
+        """
+        Generate the logit mask for the given ``dtype``.
+
+        Elements of the mask that are ``False`` are set to the
+        minimum value of the ``dtype`` and the rest to zero.
```
danieldk, line 169: We should probably add something saying that when used in softmax, that _adding_ this mask to the logits results in (near-)zero probabilities for the elements that are ``False``.

curated_transformers/layers/embeddings.py
```diff
@@ -56,10 +56,10 @@ def __init__(
 
     def forward(self, input: Tensor) -> Tensor:
         """
-        Returns the positional embedding for the input.
+        Apply sinusoidal embeddings to the input.
```
danieldk, line 59: This change is incorrect. As it is now, it only returns the positional embedding or the input.

curated_transformers/layers/embeddings.py
```diff
@@ -56,10 +56,10 @@ def __init__(
...
 
         :param input:
-            Input.
+            Input to apply the sinusoidal embeddings to.
```
danieldk, line 62: See above.

curated_transformers/layers/feedforward.py
```diff
@@ -25,15 +25,23 @@ class PointwiseFeedForward(Module):
...
     Gated Linear Units (`Dauphin et al., 2016`_; `Shazeer, 2020`_) are also
-    supported. Gating applies the transformation
-    :math:`(g(xW_g + b_g) * (xW_1 + b_1))W_2 + b_2`, where :math:`W_g` and :math:`b_g`
-    are the affine transformations for the gate.
+    supported. Gating applies the following transformation:
+
+    .. math::
+        (g(xW_g + b_g) * (xW_1 + b_1))W_2 + b_2
+
+    :math:`W_g` and :math:`b_g` are the affine transformations for the gate.
```
danieldk, line 44:
```suggestion
    :math:`W_g` and :math:`b_g` are the affine transformation for the gate.
```

(Also wrong in the original doc.)

curated_transformers/models/falcon/layer.py
```diff
@@ -118,7 +118,7 @@ def forward(
             corresponding mask element is set to ``False`` are ignored
             during attention calculation.
 
-            *Shape:* ``(batch_size, seq_len)``
+            *Shape:* ``(batch_size, 1, 1, seq_len)``
```
danieldk, line 121: Maybe we should consider removing the shape from attention mask arguments? I think the shape makes sense for `Tensor` arguments. But for `AttentionMask` the shape is already checked/enforced by the class itself. Having the shape in the argument doc suggest that this is something that the user should take care of.

curated_transformers/tokenizers/legacy/sentencepiece_tokenizer.py
```diff
@@ -20,7 +20,7 @@ def __init__(
         processor: SentencePieceProcessor,
     ):
         """
-        Construct a sentence piece tokenzier.
+        Construct a tokenzier from ``curated-tokenizers`` SentencePiece processor.
```
danieldk, line 23:
```suggestion
        Construct a tokenizer from ``curated-tokenizers`` SentencePiece processor.
```

curated_transformers/tokenizers/tokenizer.py
```diff
@@ -74,7 +74,7 @@ def padded_tensor(
         device: Optional[torch.device] = None,
     ) -> Tensor:
         """
-        Padded CPU tensor of the piece identifiers.
+        Generate a padded tensor of the piece identifiers.
```
danieldk, line 77: Maybe we should add a note here that the actual padding id typically doesn't matter when an attention mask is used, as long as the id is a valid vocabulary index? As a user I would be worried that I have to give some padding identifier, but how do I know which one?

On that note, we could consider just giving it a default value of `0` as well.

docs/source/api.rst
```diff
@@ -1,12 +1,8 @@
 API
 ===
 
-.. attention::
-   This library is currently in the tech preview stage - it is possible that models 
-   and their APIs will change in incompatible ways before ``1.0.0``.
-
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 3
```
danieldk, line 5: Too messy. A ToC should provide a quick overview of the structure of the documentation, 3 is very overwhelming.

docs/source/conf.py
```diff
@@ -38,7 +38,7 @@
     # Toc options
     "collapse_navigation": False,
     "sticky_navigation": True,
-    "navigation_depth": 3,
+    "navigation_depth": 4,
```
danieldk, line 41: In contrast to above, I think 4 is fine here. I think it's only used for the sidebar (?) which uses progressive disclosure.

curated_transformers/generation/falcon.py
shadeMe: >  If the generated documentation is not clear, we should change the documentation of the superclass constructor.

Do you have a suggestion? If we don't override the ctor here, the docstring of the `GeneratorWrapper` ctor will be displayed, but that results in an inconsistency b'ween the Falcon and Dolly v2 generators - The latter doesn't allow a default generation config to be passed to the ctor but the former will. Same with the Llama generator.

curated_transformers/layers/embeddings.py
shadeMe: Reverted.

curated_transformers/layers/embeddings.py
shadeMe: Indeed! reverted.

README.md
shadeMe: Reverted.

curated_transformers/models/falcon/layer.py
shadeMe: Fair point. Removed.

docs/source/api.rst
shadeMe: Done.

curated_transformers/tokenizers/tokenizer.py
shadeMe: Done.

curated_transformers/generation/falcon.py
danieldk: Good point, let's stick with this.

danieldk (approved):
LGTM, there are still some merge conflicts that need to be resolved.

PR #273, opened 2023-07-31
Factor out embeddings into a shared layer

danieldk:
## Description

Share the same embeddings layer in all model. While at it, improve some naming:

- `input_ids` -> `piece_ids`
- `position_ids` -> `positions`
- `token_type_embeddings` -> `type_embeddings`
- `token_type_ids` -> `type_ids`
- `word_embeddings` -> `piece_embeddings`

And make sure that positions are passed through in various places where they weren't.

### Types of change

Maintenance

## Checklist


curated_transformers/layers/transformer.py
```diff
@@ -3,12 +3,197 @@
...
+
+    embed_output_dropout: Module = Identity()
+    proj_output_dropout: Module = Identity()
+
+
+class TransformerEmbeddings(Module):
+    """
+    Transformer embeddings layer.
+
+    This is a generic transformer embedding layer. The layer alwas has piece
```
shadeMe, line 54:
```suggestion
    This is a generic transformer embedding layer. The layer always has piece
```

curated_transformers/layers/transformer.py
```diff
@@ -3,12 +3,197 @@
...
+    def __init__(
+        self,
+        *,
+        dropouts: EmbeddingsDropouts,
+        embedding_width: int,
+        hidden_width: int,
+        layer_norms: EmbeddingsLayerNorms,
+        n_pieces: int,
+        n_positions: Optional[int],
+        n_types: Optional[int],
```
shadeMe, line 68: I think we use `num_` as the prefix for counts in other places.

curated_transformers/layers/transformer.py
```diff
@@ -3,12 +3,197 @@
...
+            embeddings += position_embeddings
+
+        embeddings = self.embed_output_layer_norm(embeddings)
+        embeddings = self.embed_output_dropout(embeddings)
+
+        if self.projection is not None:
+            embeddings = self.projection(embeddings)
+            embeddings = self.proj_output_layer_norm(embeddings)
+            return self.proj_output_dropout(embeddings)
+
```
shadeMe, line 192:
```suggestion
```

curated_transformers/models/module.py
```diff
@@ -101,23 +101,31 @@ class EncoderModule(Module):
     @abstractmethod
     def forward(
         self,
-        input_ids: Tensor,
+        piece_ids: Tensor,
```
shadeMe, line 104: Probably for a different PR, but we should check and normalize where we want to pass positional arguments and where keyword-only arguments work better. I don't think we currently follow a standard protocol that spans the entire codebase.

danieldk: My main rule has been when there can be confusion between arguments because they take the same type. But we haven't really nailed that down much. E.g. if you have:

```
def foo(a: Tensor, b: Tensor):
```

then it's really easy to mix up the arguments and type checking won't help. So, we could make them keyword arguments:

```
def foo(*, a: Tensor, b: Tensor):
```

which is nice and clean. However,

```
def foo(a: Tensor, *, b: Tensor):
```

would probably be enough, since `a l is disambiguated by virtue of needing to specify `b`. However, I don's like this, _except_ in cases where it `a` would very obviously the first, most important argument (e.g. piece ids for a transformer). Especially in the case where the other arguments have default values, it would be quite jarring having to specify the name of that first argument.

I guess there is sort of another concern when types are different, that the argument list may get too long to use positional arguments. But it's hard to come up with a cut-off.

curated_transformers/layers/transformer.py
danieldk: We should probably do a consistency sweep of things across the code base before the final release, but I think I'd prefer renaming all these `num_` cases to `n_`. I think `n` is a more widely used abbreviation for number and we only have `num_` because HF model configurations have this prefix?

svlandeg: I'd personally prefer `n_` as well :-)

shadeMe: Sounds good!

shadeMe (approved):

PR #274, opened 2023-07-31
Add initial semver documentation

danieldk:
## Description

Describes the basic semver promises and discusses how API compatibility is handled in two specific cases. In the future we can add more constructs to this section as they crop up.

### Types of change

New docs.

## Checklist


shadeMe (approved):
LGTM! We just need to finalize the symbols that we want to be part of the public API contract.

PR #279, opened 2023-08-02
Expose public APIs in top-level modules

danieldk:
## Description

Expose public APIs in top-level modules. Also add `__all__` to these modules.

### Types of change

Maintenance

## Checklist


curated_transformers/layers/__init__.py
```diff
@@ -17,7 +18,37 @@
...
+    TransformerEmbeddings,
     TransformerLayerNorms,
 )
+
+__all__ = [
+    "Activation",
+    "AttentionHeads",
+    "AttentionLinearBiases",
+    "AttentionMask",
+    "CacheProtocol",
```
shadeMe, line 34: We should add this to the API docs as well.

curated_transformers/layers/__init__.py
```diff
@@ -17,7 +18,37 @@
 from .normalization import RMSNorm
 from .transformer import (
     DecoderLayer,
+    EmbeddingsDropouts,
+    EmbeddingsLayerNorms,
```
shadeMe, line 22: Wondering if `EmbeddingDropouts` and `EmbeddingLayerNorms` sound better.

curated_transformers/tests/quantization/test_generation.py
```diff
@@ -45,6 +45,7 @@ def _check_quantized_generator_output(output, expected_keywords):
     # different GPUs (which can use different kernels), we can't reliably expect the
     # output to match a string verbatim. So, we'll just look for specific, low-frequency
     # keywords as a way to detect if gibberish/irrelevant text is being generated.
+    print(output)
```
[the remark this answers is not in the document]
danieldk: Fixed. The test fails on my 3090 BTW 😢 . One or more words missing, but the output looks ok, just different.

curated_transformers/layers/__init__.py
danieldk: I'll make a separate PR, since we should also move this to `models` (it's only used by the model outputs).

curated_transformers/layers/__init__.py
danieldk: Maybe 🤔, should also be in a separate PR.

PR #281, opened 2023-08-02
Make the attention mask a mandatory argument of models

danieldk:
## Description

It is very easy to forget to pass the attention mask, giving bad output.
On the other hand it is very unlikely that someone doesn't want to use
an attention mask. In such cases, it is always possible to create a
custom mask all elements set to `True`.

While at it, also consistently make `piece_ids`/`attention_mask`
positional arguments, while making all other arguments kwarg-only.

## Checklist

### DIFF
--- a/docs/source/deployment.rst
+++ b/docs/source/deployment.rst
@@ -36,16 +36,18 @@ inputs as a tuple. For example, we can trace a decoder as follows:
 
    import torch
    import torch.jit
+   from curated_transformers.layers import AttentionMask
    from curated_transformers.models import AutoDecoder
 
    device = torch.device("cuda", index=0)
    decoder = AutoDecoder.from_hf_hub(name="tiiuae/falcon-7b", device=device)
    X_example = torch.zeros(4, 20, dtype=torch.long, device=device)
-   traced = torch.jit.trace(decoder, (X_example),))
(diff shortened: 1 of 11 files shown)


docs/source/deployment.rst
```diff
@@ -36,16 +36,18 @@ inputs as a tuple. For example, we can trace a decoder as follows:
...
-   traced = torch.jit.trace(decoder, (X_example),))
+   mask_example = AttentionMask(torch.ones((4, 20), dtype=torch.bool, device=device))
+   traced = torch.jit.trace(decoder, (X_example, mask_example))
 
-As you can see, we are feeding the model with an all-zeros tensor during
-tracing. This is not really an issue, as long as the inputs allows the model to
-run normally, tracing can do its work.
+As you can see, we are feeding the model with an all-zeros piece identifier and
+all-ones mask tensors during tracing. This is not really an issue, as long as
+the inputs allows the model to run normally, tracing can do its work.
```
shadeMe, line 50:
```suggestion
As you can see, we are feeding the model with an all-zeros piece identifier tensor and
an all-ones mask tensor during tracing. This is not really an issue - as long as
the inputs allow the model to run normally, tracing can do its work.
```

shadeMe (approved):
LGTM!

PR #296, opened 2023-08-09
Add `MPTGenerator`

danieldk:
## Description

Add the `MPTGenerator` class. This is pretty straightforward, except that I had
to add a workaround because we don't support weight tying yet.

### Types of change

Feature.

## Checklist


shadeMe (approved):
LGTM! One minor change; can be merged afterwards.

PR #309, opened 2023-08-14
Fix import paths to use a top-level import

shadeMe:
## Description

There were a bunch of imports in the docs that didn't use the top-(module-)level path. This PR fixes that. Plus a bunch of other docfixes.

## Checklist


svlandeg (approved):
LGTM!

PR #310, opened 2023-08-15
Add support for `safetensors` checkpoints

shadeMe:
## Description

This PR introduces support for model checkpoints that saved in the `safetensors` serialization format. `FromHFHub` mixins will prefer to initialize models from `safetensors` checkpoints if the following conditions are met:
- The `safetensors` library is installed.
- The model repo contains `.safetensors` checkpoint files.

If one of the above conditions are not met, the loader will fallback to using regular PyTorch checkpoints. This also applies to sharded checkpoints.

Additionally, the user can force the usage of a specific checkpoint type with the `use_model_checkpoint_type` context manager. 

## Checklist

- [X] I confirm that I have the right to submit this contribution under the project's MIT license.

### DIFF
--- a/curated_transformers/util/hf.py
+++ b/curated_transformers/util/hf.py
@@ -1,13 +1,18 @@
 import json
 import warnings
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import huggingface_hub
 from requests import HTTPError, ReadTimeout  # type: ignore
 
+from .._compat import has_safetensors
+from .serde import MODEL_CHECKPOINT_TYPE, ModelCheckpointType
+
 HF_MODEL_CONFIG = "config.json"
 HF_MODEL_CHECKPOINT = "pytorch_model.bin"
+HF_MODEL_CHECKPOINT_SAFETENSORS = "model.safetensors"
 HF_MODEL_SHARDED_CHECKPOINT_INDEX = "pytorch_model.bin.index.json"
+HF_MODEL_SHARDED_CHECKPOINT_INDEX_SAFETENSORS = "model.safetensors.index.json"
 HF_MODEL_SHARDED_CHECKPOINT_INDEX_WEIGHTS_KEY = "weight_map"
 HF_TOKENIZER_CONFIG = "tokenizer_config.json"
 SPECIAL_TOKENS_MAP = "special_tokens_map.json"
@@ -73,11 +78,13 @@ def get_model_config_filepath(name: str, revision: str) -> str:
         )
 
 
-def get_model_checkpoint_filepaths(name: str, revision: str) -> List[str]:
+def get_model_checkpoint_filepaths(
+    name: str, revision: str
+) -> Tuple[List[str], ModelCheckpointType]:
     """
-    Return a list of local file paths to PyTorch checkpoints that belong
-    to the Hugging Face model. In case of non-sharded models, a single file path
-    is returned. In case of sharded models, multiple file paths are returned.
+    Return a list of local file paths to  checkpoints that belong to the Hugging
+    Face model. In case of non-sharded models, a single file path is returned. In
+    case of sharded models, multiple file paths are returned.
 
     If the checkpoints are not found in the cache, they are downloaded from
     Hugging Face Hub.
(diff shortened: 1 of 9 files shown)


curated_transformers/util/serde.py
```diff
@@ -18,10 +34,52 @@
 ]
 
 
+class ModelCheckpointType(Enum):
+    """
+    Types of model checkpoints supported by Curated Transformers.
+    """
+
+    #: PyTorch checkpoint.
+    PYTORCH = 1
```
danieldk, line 43: Maybe we should be more specific here, since it is possible that we have other types in the future? Maybe `PYTORCH _LOAD` or `PYTORCH_STATE_DICT`? Also, I'd start numbering at 0, C-style.

curated_transformers/util/serde.py
```diff
@@ -18,10 +34,52 @@
...
+
+    #: Hugging Face `Safetensors <https://github.com/huggingface/safetensors>`_ checkpoint.
+    SAFETENSORS = 2
```
danieldk, line 46: They call them `SafeTensors` in the Rust source, so maybe a good excuse to add an underscore to the name here?

curated_transformers/util/serde.py
```diff
@@ -18,10 +34,52 @@
...
+
+
+# When `None`, behaviour is implementation-specific.
```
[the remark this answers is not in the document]
danieldk: I wouldn't do this at all and just let the library decide how to load. We can always add more control when it turns out to be necessary, but if we add this now, there is no way back.

curated_transformers/util/serde.py
```diff
@@ -33,16 +91,23 @@ def load_model_from_checkpoints(
...
-        `curated-transformers` format.
+        ``curated-transformers`` format.
     :param tensor_to_param_converter:
         Callback to perform custom conversions of the loaded parameters.
         Useful for loading quantized weights.
     :param device:
         Device in which to place the loaded parameters.
     """
-    state_dicts = _load_state_dicts_from_checkpoints(filepaths)
+    checkpoint_type_to_loader = {
```
danieldk, line 105: I think it's cleaner if we add this as a method (even if private) to the enum.

docs/source/utils.rst
```diff
@@ -1,11 +1,19 @@
 Utilities
 =========
 
+Serialization
+-------------
+
+.. autoclass:: curated_transformers.util.ModelCheckpointType
```
danieldk, line 7: I wouldn't make this public API yet.

docs/source/utils.rst
```diff
@@ -1,11 +1,19 @@
...
+   :members:
+
 Context Managers
 ----------------
 
 .. autofunction:: curated_transformers.layers.attention.enable_torch_sdp
 
+.. autofunction:: curated_transformers.util.use_model_checkpoint_type
```
danieldk, line 15: Same for this, see comment above.

danieldk (commented):
I probably need to do another pass when I feel a bit better, but here are some initial comments.

curated_transformers/util/serde.py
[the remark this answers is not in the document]
shadeMe: > I wouldn't do this at all and just let the library decide how to load.

I'm not sure I follow - are you referring to `safetensors` when you say _library_ here? 

Maybe that comment is confusing, but if you're saying _let `curated-transformers` decide how to load_, that's how it's currently implemented. AFAICT, the `safetensors` library doesn't contain any logic to select the checkpoint files - the API just takes paths to `.safetensors` files and returns a `state_dict`.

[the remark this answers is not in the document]
danieldk: > Maybe that comment is confusing, but if you're saying let curated-transformers decide how to load, that's how it's currently implemented.

Currently the user can also decide how to load with the context manager.

curated_transformers/util/serde.py
shadeMe: Renamed to `PYTORCH_STATE_DICT` and renumbered.

curated_transformers/util/serde.py
shadeMe: Done.

curated_transformers/util/serde.py
shadeMe: Moved to the enum.

docs/source/utils.rst
shadeMe: Removed.

docs/source/utils.rst
shadeMe: Removed.

shadeMe:
Looks like a regression in HF `transformers` 4.32.0.

curated_transformers/util/hf.py
```diff
@@ -87,50 +94,98 @@ def get_model_checkpoint_filepaths(name: str, revision: str) -> List[str]:
...
-        raise ValueError(
-            f"Couldn't find a valid PyTorch checkpoint for model `{name}` "
-            f"(revision `{revision}`) on HuggingFace Model Hub"
+    def get_checkpoint_paths(
+        checkpoint_type: ModelCheckpointType,
+    ) -> List[str]:
+        checkpoint_type_to_string = {
+            ModelCheckpointType.PYTORCH_STATE_DICT: "PyTorch",
+            ModelCheckpointType.SAFE_TENSORS: "Safetensors",
+        }
```
danieldk, line 107: Maybe a method on the enum could return the nice names? Then we don't need an auxiliary mapping.

curated_transformers/util/hf.py
```diff
@@ -87,50 +94,98 @@ def get_model_checkpoint_filepaths(name: str, revision: str) -> List[str]:
...
+
+        primary_checkpoint_filenames = {
+            ModelCheckpointType.PYTORCH_STATE_DICT: HF_MODEL_CHECKPOINT,
+            ModelCheckpointType.SAFE_TENSORS: HF_MODEL_CHECKPOINT_SAFETENSORS,
+        }
+        sharded_checkpoint_index_filenames = {
+            ModelCheckpointType.PYTORCH_STATE_DICT: HF_MODEL_SHARDED_CHECKPOINT_INDEX,
+            ModelCheckpointType.SAFE_TENSORS: HF_MODEL_SHARDED_CHECKPOINT_INDEX_SAFETENSORS,
+        }
```
danieldk, line 116: Maybe these should be constants outside this function?

curated_transformers/util/serde.py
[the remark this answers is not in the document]
danieldk: Yeah, but I'd underscore the context var and context manager for now to make it very clear that this is not for public consumption.

curated_transformers/util/hf.py
shadeMe: Added.

curated_transformers/util/hf.py
shadeMe: Moved.

curated_transformers/util/serde.py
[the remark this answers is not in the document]
shadeMe: Done.

danieldk (approved):

PR #313, opened 2023-08-28
Module and parameter sharing

shadeMe:
## Description

This PR introduces the `Shareable` mixin which can be used to share parameters between modules inside a model. Sharing can be applied at both the module-level (entire `Module`s can be shared between layers) and the parameter-level (individual parameters can be shared between different `Module`s). 

The implementation takes place in following stages:
- After the model has been constructed and its structure has been finalized (any fix-ups to the model graph that requires changes to the model structure like quantization-related `Module` swaps, etc), the loading code in `FromHFHub` will initialize the shareable data by replacing target modules and parameters with placeholders.
- Then, the checkpoints are loaded and the parameters and copied into the model. Any missing parameters that pertain to shared data (e.g: the `weight` parameter of a causal LM's `output_embeddings` layer that is shared with the `weight`s of the `piece_embeddings`) are ignored.
- Once the checkpoints are loaded, the `tie_shared_data` method is called on the `Shareable` model, which performs the sharing logic.

This implementation allows for minimal changes to the actual model code. The mixin also provides methods to untie and clone the shared parameters to facilitate serialization.

Additionally, the `MPTCausalLM` has been modified to use parameter-level sharing for its output embeddings. This should be a backwards-compatible change.

## Checklist


shadeMe:
Reverting to draft. This proposal will be revisited after v2.0.0.

PR #317, opened 2023-09-02
bug: use MPTCausalLM for MPT instead of LlamaCausalLM  

mayankjobanputra:
I think the MPT Generator should be using MPTCausalLM instead of LlamaCausalLM

## Checklist


danieldk:
Thanks! It was not an issue in normal cases, since the loader passes an MPT generator, but this does fix the types.

danieldk (approved):

PR #328, opened 2023-09-15
Persist model config to aid serialization

shadeMe:
## Description

This PR introduces a top-level `TransformerModule` class in the inheritance hierarchy that stores the model config and exposes a property to query the same. The implementation is generic over the config type, so subclasses return the instance with the concrete type.

## Checklist

### DIFF
--- a/docs/source/causal-lm.rst
+++ b/docs/source/causal-lm.rst
@@ -7,10 +7,12 @@ Base Classes
 .. autoclass:: curated_transformers.models.CausalLMModule
    :members:
    :show-inheritance:
+   :inherited-members: Module
 
 .. autoclass:: curated_transformers.models.TransformerCausalLM
    :members:
    :show-inheritance:
+   :inherited-members: Module
 
 Architectures
 -------------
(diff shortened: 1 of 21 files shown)


docs/source/causal-lm.rst
```diff
@@ -7,10 +7,12 @@ Base Classes
 .. autoclass:: curated_transformers.models.CausalLMModule
    :members:
    :show-inheritance:
+   :inherited-members: Module
```
danieldk, line 10: Doesn't this add a lot of methods from PyTorch? Maybe we want `TransformerModule` here?

danieldk (commented):
Great to have this. One question besides the `Any` typing discussed on Slack.

docs/source/causal-lm.rst
shadeMe: My understanding of the [RST docs](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) is that the classes that follow the directive are _excluded_ from the member list? At least it works as intended locally (from my testing).

danieldk (approved):

PR #330, opened 2023-09-18
Add a default `dtype` to model configs

shadeMe:
## Description

This lets us fallback to a known/expected datatype if the HF model config doesn't contain that metadata.

## Checklist

### DIFF
--- a/curated_transformers/models/module.py
+++ b/curated_transformers/models/module.py
@@ -1,14 +1,14 @@
 from abc import abstractmethod
-from typing import Generic, List, Optional, TypeVar, Union
+from typing import Generic, List, Optional, TypeVar
 
 from torch import Tensor
 from torch.nn import Module
 
 from ..layers.attention import AttentionMask
-from .config import ConfigDataclass, TransformerConfig
+from .config import TransformerConfig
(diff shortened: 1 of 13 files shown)


curated_transformers/models/module.py
```diff
@@ -1,14 +1,14 @@
...
 from torch import Tensor
 from torch.nn import Module
 
 from ..layers.attention import AttentionMask
-from .config import ConfigDataclass
+from .config import ConfigDataclass, TransformerConfig
 from .output import CacheT, CausalLMOutputWithCache, ModelOutput, ModelOutputWithCache
 
-ConfigT = TypeVar("ConfigT", bound=ConfigDataclass)
+ConfigT = TypeVar("ConfigT", bound=Union[ConfigDataclass, TransformerConfig])
```
danieldk, line 11: If we have a superclass for everything now anyway, I think we could just make `TransformerConfig` a data class. At any rate, I don't think we'd want to use the union here, because the `ConfigDataclass` bound weakens the typing. The typevar could still be useful, because it allows the non-auto classes to return the concrete type.

shadeMe: I've restricted the bound to `TransformerConfig` (which is already a dataclass).

danieldk (approved):

PR #332, opened 2023-09-26
Add support for converting Curated Transfomer state dicts Hugging Face compatible state dicts

shadeMe:
## Description

This is the first of a series of PRs that aim to implement serialization support by way of producing serialized models that are usable with the Hugging Face `transformers` library.

The main addition of this PR is a selection of string transformation classes (in `util.string`) that are used to perform reversible transformations of parameter names in PyTorch state dicts. The existing ad-hoc transformation code has been rewritten with the above classes, and the Curated-to-HF conversions are tested for all supported models.

Additionally, the module structure has been changed in the following manner: The `models.hf_hub` module has been split into sub-modules `models.hf_hub.conversion` and `models.hf_hub.mixin`. The `FromHFHub` mixin is re-exported in `models.hf_hub` to preserve backward compatibility.

## Checklist

### DIFF
--- a/curated_transformers/models/albert/_hf.py
+++ b/curated_transformers/models/albert/_hf.py
@@ -18,19 +18,15 @@ HF_PARAM_KEY_TRANSFORMS: List[StringTransform] = [
         (r"^albert_layer_groups\.", "groups."), (r"^groups\.", "albert_layer_groups.")
     ),
     # Inner layers.
-    StringTransformations.regex_sub_invertible((".albert_layers.", ".group_layers.")),
+    StringTransformations.sub(".albert_layers.", ".group_layers."),
     # Attention blocks.
-    StringTransformations.regex_sub_invertible((".attention.", ".mha.")),
-    StringTransformations.regex_sub_invertible(
-        (".mha.LayerNorm", ".attn_residual_layer_norm")
-    ),
-    StringTransformations.regex_sub_invertible((".mha.dense", ".mha.output")),
+    StringTransformations.sub(".attention.", ".mha."),
+    StringTransformations.sub(".mha.LayerNorm", ".attn_residual_layer_norm"),
+    StringTransformations.sub(".mha.dense", ".mha.output"),
     # Pointwise feed-forward layers.
-    StringTransformations.regex_sub_invertible((".ffn.", ".ffn.intermediate.")),
-    StringTransformations.regex_sub_invertible((".ffn_output.", ".ffn.output.")),
-    StringTransformations.regex_sub_invertible(
-        (".full_layer_layer_norm.", ".ffn_residual_layer_norm.")
-    ),
+    StringTransformations.sub(".ffn.", ".ffn.intermediate."),
+    StringTransformations.sub(".ffn_output.", ".ffn.output."),
+    StringTransformations.sub(".full_layer_layer_norm.", ".ffn_residual_layer_norm."),
     # Embeddings.
     StringTransformations.replace(
         "embeddings.word_embeddings.weight", "embeddings.piece_embeddings.weight"
--- a/curated_transformers/models/albert/_hf.py
+++ b/curated_transformers/models/albert/_hf.py
@@ -1,45 +1,57 @@
 from typing import Any, Callable, Dict, List, Tuple, Union
 
 from ...layers.activations import Activation
-from ...util.string import StringTransform, StrLStrip, StrRepl, StrSub, StrSubInv
+from ...util.string import (
+    StringRemovePrefix,
+    StringReplace,
+    StringSubInvertible,
+    StringSubRegEx,
+    StringTransform,
+)
 from ..hf_hub.conversion import process_hf_keys
 from .config import ALBERTConfig
 
 # Order-dependent.
 HF_PARAM_KEY_TRANSFORMS: List[StringTransform] = [
     # Prefixes.
-    StrLStrip("albert.", reversible=False),
-    StrSub(
+    StringRemovePrefix("albert.", reversible=False),
+    StringSubRegEx(
         (r"^encoder\.(embedding_|albert_layer)", "\\1"),
         (r"^(embedding_|albert_layer)", "encoder.\\1"),
     ),
     # Layer groups
-    StrSub(
+    StringSubRegEx(
         (r"^albert_layer_groups\.", "groups."), (r"^groups\.", "albert_layer_groups.")
     ),
     # Inner layers.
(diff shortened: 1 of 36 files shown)


curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+        super().__init__(backward is not None)
+
+        self.forward = forward
+        self.backward = backward
+
+    def _apply(self, string: str) -> str:
+        return re.sub(self.forward[0], self.forward[1], string)
+
+    def _revert(self, string: str) -> str:
+        assert self.backward is not None
```
danieldk, line 80: This should probably be a regular exception. It's not some internal failure, but downstream misuse of the interface.

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+        return re.sub(self.backward[0], self.backward[1], string)
+
+
+class StrSubInv(StrSub):
```
danieldk, line 84: I'd say something like:

```suggestion
class StringSubRevesible(StrSub):
```

`Inv` is not really clear.

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+            and the second the replacement, neither of which
+            can contain RegEx meta-characters.
+        """
+        super().__init__(
+            forward=(f"{re.escape(forward[0])}", forward[1]),
+            backward=(f"{re.escape(forward[1])}", forward[0]),
+        )
+
+
+class StrRepl(StrSub):
```
danieldk, line 105: What's the difference with `StrSub`? If the only the difference is that this does not use regular expressions, `StrSub` should probably have _regex_ in its name. And since this matches the whole string, it should probably have something in its name that indicates that.

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, prefix: str, *, reversible: bool = True):
+        """
+        Construct a reversible left strip.
+
+        :param prefix:
+            Prefix to be stripped.
+        """
+        super().__init__(
+            forward=(f"^{re.escape(prefix)}", ""),
```
danieldk, line 138: Add a TODO that this should be replaced by [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) in Python 3.9.

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+            backward=(r"^(.)", f"{prefix}\\1") if reversible else None,
```
danieldk, line 139: Can't we concatenate the prefix rather than the extra overhead of going through regular expressions?

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+        Construct a reversible replacement.
+
+        :param replacee:
+            The full string to be replaced.
+        :param replacement:
+            The replacement string.
+        """
+        super().__init__(
+            forward=(f"^{re.escape(replacee)}$", replacement),
+            backward=(f"^{re.escape(replacement)}$", replacee) if reversible else None,
```
danieldk, line 121: Unnecessary use of `re`, match the string using `==`. Easier to read and faster.

curated_transformers/models/albert/_hf.py
```diff
@@ -1,25 +1,46 @@
...
-        "embeddings.LayerNorm.bias": "embeddings.embed_output_layer_norm.bias",
-        # Embedding projection
-        "encoder.embedding_hidden_mapping_in.weight": "embeddings.projection.weight",
-        "encoder.embedding_hidden_mapping_in.bias": "embeddings.projection.bias",
-    }
-)
+# Order-dependent.
+HF_PARAM_KEY_TRANSFORMS: List[StringTransform] = [
+    # Prefixes.
+    StrLStrip("albert.", reversible=False),
```
danieldk, line 11: I wonder if this is a case where a nested class with factory methods works better, so we could have something like:

```
    StringTransform.remove_prefix("albert.", reversible=False),
```

Or to make it clear that a structure is built:

```
    StringTransform.RemovePrefix("albert.", reversible=False),
```

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+
+class StrLStrip(StrSub):
```
danieldk, line 125:
```suggestion
class StringRemovePrefix(StrSub):
```

lstrip is a bit confusing, because Python `lstrip` removes using a set of characters.

curated_transformers/models/albert/_hf.py
shadeMe: Good idea! I like the factory method, given that we already have a base class. Will rewrite the implementation.

curated_transformers/util/string.py
shadeMe: Renamed to `StringSubInvertible` (since using reversible might be confusing since all transforms are allowed to be reversible).

curated_transformers/util/string.py
```diff
@@ -0,0 +1,140 @@
...
+        :returns:
+            Reverted string.
+        """
+        if self._reversible:
+            return self._revert(string)
+        else:
+            return string
+
+
+class StrSub(StringTransform):
```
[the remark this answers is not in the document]
shadeMe: Renamed to `StringSubRegEx`.

curated_transformers/util/string.py
shadeMe: Fixed.

curated_transformers/util/string.py
shadeMe: Renamed to `StringReplace`.

curated_transformers/util/string.py
shadeMe: Done.

curated_transformers/util/string.py
shadeMe: Replaced the backward operation with a string concat.

curated_transformers/util/string.py
shadeMe: Rewritten to remove the use of  `re`.

curated_transformers/util/string.py
shadeMe: Renamed.

curated_transformers/models/albert/_hf.py
shadeMe: Added `StringTransformations` to expose the factory methods.

danieldk (approved):
Nice 🎉 , this makes the conversions so much cleaner.

PR #333, opened 2023-09-27
Add support for converting Curated Transformer configs to Hugging Face compatible configs

shadeMe:
## Description

This PR follows up on #332, adding support for bidirectional conversions of model configs between Curated Transformers and Hugging Face. This is facilitated by the following classes:
* `HFConfigKey` - Descriptor for a HF model config key. Defines how its mapped to the CT config.
* `HFConfigKeyDefault` - Wrapper around a default value for a HF config key that allows for the generic handling of optional keys.
* `HFSpecificConfig` - A set of hardcoded keys that are required to be part of every HF model config. Overridden on a per-model basis and merged into the final config dictionary.
* `CommonHFKeys` and `CommonCuratedToHFConverters` - Shared config key descriptors and conversion functions.

The `FromHFHub` mix has been expanded to provide methods for the conversion of the configs.

All currently supported models except Falcon support bidirectional config conversions implicitly. The conversion of the Falcon model config is more complicated as we support loading from two different model implementations - `RefinedWebModel` and `Falcon`. That and the complicated `new_decoder_architecture` situation makes it difficult to allow the full range of conversion. Since the latter implementation is now in the mainline `transformers` branch, we'll do the following:
* Conversion from both RWM and Falcon HF architectures is fully supported.
* The CT Falcon config will only be converted to the mainline Falcon HF config/architecture.

## Checklist


curated_transformers/models/hf_hub/conversion.py
```diff
@@ -1,15 +1,278 @@
...
+        Removes the key the given keyword argument dictionary.
+
+        :param kwargs:
+            Keyword argument dictionary.
+        """
+        kwargs.pop(self.curated_config_kwarg)
+
+
+@dataclass
+class HFConfigKeyDefault:
```
rmitsch, line 132: Why a class for this?

curated_transformers/models/hf_hub/mixin.py
```diff
@@ -221,6 +235,7 @@ def from_repo(
         model_repo = ModelRepository(repo)
         config = model_repo.model_config()
         model = cls.from_hf_config(hf_config=config, device=torch.device("meta"))
+        assert isinstance(model, Module)
 
         # Convert the model to the expected dtype.
         assert isinstance(model, TransformerModule)
```
rmitsch, line 241: Is it a possibility that this fails? Otherwise you could also use I would still like to know if it’s possible to annotate unlabelled spans in Prodigy and how., no?

rmitsch (commented):
I'm taking a superficial pass through the code, I don't want to keep you too long (and I won't be able to appreciate the details at this point anyway due to my lack of experience with `curated-transformers`).

I'll finish the review tomorrow.

curated_transformers/models/hf_hub/conversion.py
shadeMe: The presence of a default value indicates whether a particular HF config key is optional. We could simply use an `Optional[Any]` type to encode the default value generically, but some keys might use `None` itself as a default value. So, we need a wrapper type to disambiguate such cases.

curated_transformers/models/hf_hub/mixin.py
shadeMe: Looks like an errant copy-paste error in the above comment. But no, the output of `from_hf_config` is guaranteed to be a subclass of `torch.nn.Module`. That assertion is to satisfy `mypy` (we call a `Module` method a bit later, and it would complain about it otherwise).

rmitsch: Copy-paste error indeed, sorry. Fixed it. My suggestion with `typing.cast` should work here then, IMO (but it's fine to leave it as-is).

shadeMe: Generally speaking, we've endeavored to avoid casting types in this codebase. We try to explicitly encode the types and let `mypy` do the type checking. The only exception is when the latter cannot correctly infer the type like with mix-in code above. Prior to this PR, the mix-in class had an abstract `to` method to workaround this, but that formulation didn't work with with generics. So, we went with the assertion which is preferable over a cast as it'll raise an error at the callsite if the precondition fails.

curated_transformers/models/mpt/causal_lm.py
```diff
@@ -12,15 +12,15 @@
...
 from .decoder import MPTDecoder
 
 # Only provided as typing.Self in Python 3.11+.
 Self = TypeVar("Self", bound="MPTCausalLM")
 
 
-class MPTCausalLM(TransformerCausalLM[MPTConfig], FromHFHub, Quantizable):
+class MPTCausalLM(TransformerCausalLM[MPTConfig], FromHFHub[MPTConfig], Quantizable):
     """
     `MosaicML MPT`_ causal language model.
```
rmitsch, line 25: Is the `_` a typo?

rmitsch (approved):
As mentioned I can't really give valuable feedback here, but this seems alright to me and seems to be mostly a refactoring PR anyway. LGTM.

curated_transformers/models/mpt/causal_lm.py
shadeMe: Nope, that's just weird RST syntax used for inline links.

PR #349, opened 2023-10-04
Add `AttentionScorer` abstraction

shadeMe:
## Description

This PR abstracts the attention scoring mechanism, adding support for plugging in different implementations such as the disentangled attention component used in DeBERTa.

We also move the Torch SDP code into `ScaledDotProductAttention` and emit a warning when it's enabled without the correct scorer.

## Checklist


curated_transformers/layers/attention.py
```diff
@@ -877,34 +909,22 @@ def forward(
...
-        else:
-            attn = self.attention(
-                query=query,
-                key=key,
-                value=value,
-                attention_mask=combined_mask,
+        if _TORCH_SDP.get() and not isinstance(
+            self.attention_scorer, ScaledDotProductAttention
+        ):
+            warn_msg = (
```
danieldk, line 915: Not sure if we want this warning. People could be running multiple models and have them all scoped in the context manager to enable Torch SDP for all supported models.

Also, someone could make another scorer class that also uses Torch SDP.

shadeMe: Good point. I'll remove it.

danieldk (approved):

PR #351, opened 2023-10-05
Register models using `catalogue`

danieldk:
## Description

So far we have hardcoded the available encoders/decoders/causal LMs in the auto classes. This has the downside that the auto classes only work with models that are provided by Curated Transformers.

This change adds registries for encoders/decoders/causal LMs. The auto classes query the relevant registry and check which registered model supports the downloaded model (through the `hf_model_types` method of the `FromHFHub` mixin). This makes it possible to register external models with Curated Transformers, so that they can also be used with the auto classes.

Adding registries for tokenizers and generators is deferred to future PRs.

## Checklist

### DIFF
--- a/curated_transformers/models/auto_model.py
+++ b/curated_transformers/models/auto_model.py
@@ -39,7 +39,7 @@ class AutoModel(ABC, Generic[ModelT]):
         for entrypoint, module_cls in cls._registry.get_entry_points().items():
             if not issubclass(module_cls, FromHFHub):
                 warnings.warn(
-                    f"Entry point {entrypoint} cannot load from Huggingface Hub "
+                    f"Entry point `{entrypoint}` cannot load from Hugging Face Hub "
                     "since the FromHFHub mixin is not implemented"
                 )
                 continue
--- a/curated_transformers/models/auto_model.py
+++ b/curated_transformers/models/auto_model.py
@@ -1,27 +1,21 @@
+import warnings
 from abc import ABC, abstractmethod
-from typing import Dict, Generic, Optional, Type, TypeVar
+from typing import Generic, Optional, Type, TypeVar
 
 import torch
+from catalogue import Registry
 from fsspec import AbstractFileSystem
+
+from curated_transformers import registry
 
 from ..layers.cache import KeyValueCache
 from ..quantization.bnb.config import BitsAndBytesConfig
 from ..repository.fsspec import FsspecArgs, FsspecRepository
 from ..repository.hf_hub import HfHubRepository
 from ..repository.repository import ModelRepository, Repository
-from .albert import ALBERTEncoder
-from .bert import BERTEncoder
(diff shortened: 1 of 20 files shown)


curated_transformers/models/auto_model.py
```diff
@@ -32,22 +26,33 @@ class AutoModel(ABC, Generic[ModelT]):
...
-                f"Supported model types: {tuple(cls._hf_model_type_to_curated.keys())}"
-            )
-        assert issubclass(module_cls, FromHFHub)
-        return module_cls
+
+        supported_model_types = set()
+        for entrypoint, module_cls in cls._registry.get_entry_points().items():
+            if not issubclass(module_cls, FromHFHub):
+                warnings.warn(
+                    f"Entry point {entrypoint} cannot load from Huggingface Hub "
```
shadeMe, line 42:
```suggestion
                    f"Entry point `{entrypoint}` cannot load from Hugging Face Hub "
```

docs/source/registries.rst
```diff
@@ -0,0 +1,24 @@
+Registries
+==========
+
+All models in Curated Transformers are added to a registry. Each auto class uses
```
shadeMe, line 4: Do we want to add a comment mentioning what happens when more than one model registers for the same model type (the one that was registered first gets picked)? Or just implicitly leave that as an impl detail?

docs/source/registries.rst
```diff
@@ -0,0 +1,24 @@
...
+a registry to query which models are available. This mechanism allows
+third-party models to hook into the auto classes. This makes it possible to use
+construction methods such as ``AutoModel.from_hf_hub`` with third-party models.
```
shadeMe, line 7: Could we make `from_hf_hub` a hyperlink?

docs/source/registries.rst
danieldk: It crossed my mind, but I'm not sure what happens now and what behavior we would prefer. So maybe keep it underspecified for now until we know what we want?

shadeMe: I think we just iterate over the backing dictionary in insertion order. But keeping it under specified for now is fine, I think.

danieldk: That still doesn't tell us anything, because the order in which the entry points are inserted in the dictionary is determined by the order in which `importlib.metadata` discovers packages. This is probably in turn determined by `sys.path`. I haven't looked into how this is implemented, but if that's the case the order is undetermined, because the user controls it.

docs/source/registries.rst
danieldk: We can't because we don't expose `AutoModel` in the API docs, but maybe we should link to a concrete implementation here, is does say _such as_ after all.

danieldk: Linked.

shadeMe (approved):

PR #352, opened 2023-10-05
AutoModel: let models check if the configuration is supported

danieldk:
## Description

This will allow us to split up more complex models like Falcon into multiple classes and corresponding entry points.

## Checklist

### DIFF
--- a/curated_transformers/models/hf_hub/mixin.py
+++ b/curated_transformers/models/hf_hub/mixin.py
@@ -176,7 +176,7 @@ class FromHFHub(ABC):
         class.
 
         :param config:
-            Model configuration.
+            Hugging Face model configuration.
         :returns:
             Whether the model is supported by this class.
         """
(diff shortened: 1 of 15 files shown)


curated_transformers/models/hf_hub/mixin.py
```diff
@@ -170,14 +170,17 @@ def from_hf_hub(
...
     @abstractmethod
-    def hf_model_types(cls: Type[Self]) -> Tuple[str, ...]:
+    def is_supported(cls: Type[Self], config: Dict[str, Any]) -> bool:
         """
-        Get the Hugging Face model types supported by this model.
+        Check if the model with the given configuration is supported by this
+        class.
 
+        :param config:
+            Model configuration.
```
shadeMe, line 179:
```suggestion
            Hugging Face model configuration.
```

shadeMe (approved):
Just one minor change.

PR #354, opened 2023-10-06
Add support for file write/upload operations with `HfHubRepository`

shadeMe:
## Description

This PR introduces support for uploading files to Hugging Face Hub repositories through the current `RepositoryFile` API. In the case of `fsspec` repositories, we support implicit streaming of data using buffered IO. However, this is not possible with HF repositories due its usage of Git as a backing store. So, we need to allocate and write to files on the host's local storage device before uploading them as a Git commit operation to the remote repo.

The current implementation hides this detail by using a proxy to represent the remote file that we want to write to. The `HfHubFile` class wraps either a locally cached file from a HF repo or a temporary file on the local storage. The user can open and write to the latter like they would with any (`fsspec`) file - its contents will be uploaded to the repo as soon as the file handle is closed.

Furthermore, we also introduce the concept of transactional file writes using a context manager. This lets the user batch multiple file operations that get uploaded to the repo as a single commit. The `fsspec` implementation of transactions will be implemented in a follow-up PR.

Other changes:
- `Repository.file` returns a lazily-loaded file, i.e., its existence is only checked when the file is opened using the `open` method (or when the `exists` method is called).
- Fixed a bug in `FsspecRepository.open` that called `unstrip_protocol` on a `None` object.

## Checklist

### DIFF
--- a/curated_transformers/repository/file.py
+++ b/curated_transformers/repository/file.py
@@ -1,3 +1,4 @@
+import os
 from abc import ABC, abstractmethod
 from typing import IO, Optional
 
@@ -22,6 +23,8 @@ class RepositoryFile(ABC):
             Encoding to use when the file is opened as text.
         :returns:
             An I/O stream.
+        :raises FileNotFoundError:
+            When the file cannot be found.
         :raises OSError:
             When the file cannot be opened.
         """
@@ -37,6 +40,14 @@ class RepositoryFile(ABC):
             The repository file. If the file is not available as a local
             path, the value of this property is ``None``. In these cases
             ``open`` can be used to get the file as a file-like object.
+        """
+        ...
+
+    @abstractmethod
+    def exists(self) -> bool:
+        """
+        Returns if the file exists. This can cause the file
+        to be cached locally.
         """
         ...
 
@@ -63,3 +74,6 @@ class LocalFile(RepositoryFile):
     @property
     def path(self) -> Optional[str]:
         return self._path
+
+    def exists(self) -> bool:
+        return os.path.isfile(self._path)
--- a/curated_transformers/repository/fsspec.py
+++ b/curated_transformers/repository/fsspec.py
@@ -1,3 +1,4 @@
+import warnings
 from dataclasses import dataclass
 from typing import IO, Any, Dict, Optional
 
@@ -145,6 +146,7 @@ class FsspecTransactionContext(TransactionContext):
         return self._repo
 
     def __enter__(self):
+        warnings.warn("Transaction support is currently not implemented for fsspec")
         return self
 
     def __exit__(self, exc_type, exc_value, exc_tb):
--- a/curated_transformers/repository/fsspec.py
+++ b/curated_transformers/repository/fsspec.py
@@ -4,6 +4,7 @@ from typing import IO, Any, Dict, Optional
 from fsspec import AbstractFileSystem
 
 from .repository import Repository, RepositoryFile
+from .transaction import TransactionContext
 
 
 @dataclass
@@ -55,18 +56,26 @@ class FsspecFile(RepositoryFile):
         self._fsspec_args = FsspecArgs() if fsspec_args is None else fsspec_args
 
     def open(self, mode: str = "rb", encoding: Optional[str] = None) -> IO:
+        if ("r" in mode or "+" in mode) and not self.exists():
+            raise FileNotFoundError(
+                f"Cannot find fsspec path {self._fs.unstrip_protocol(self._path)}"
+            )
+
         try:
             return self._fs.open(
                 self._path, mode=mode, encoding=encoding, **self._fsspec_args.kwargs
             )
         except Exception as e:
             raise OSError(
-                f"Cannot open fsspec path {self._fs.unstrip_protocol(self.path)}"
-            )
+                f"Cannot open fsspec path {self._fs.unstrip_protocol(self._path)}"
+            ) from e
 
     @property
     def path(self) -> Optional[str]:
         return None
+
+    def exists(self) -> bool:
+        return self._fs.exists(self._path, **self._fsspec_args.kwargs)
 
 
 class FsspecRepository(Repository):
@@ -97,8 +106,6 @@ class FsspecRepository(Repository):
 
     def file(self, path: str) -> RepositoryFile:
         full_path = f"{self.repo_path}/{path}"
-        if not self.fs.exists(full_path, **self.fsspec_args.kwargs):
-            raise FileNotFoundError(f"Cannot find file in repository: {path}")
         return FsspecFile(self.fs, full_path, self.fsspec_args)
 
     def pretty_path(self, path: Optional[str] = None) -> str:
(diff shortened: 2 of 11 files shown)


curated_transformers/repository/hf_hub.py
```diff
@@ -31,26 +112,201 @@ def __init__(self, name: str, *, revision: str = "main"):
...
+            operations=commit_ops,
+            commit_message=commit_message,
+            run_as_future=False,
+        )
+
+    def _release_temp_files(self):
+        for temp_file in self._file_mappings.values():
+            if not temp_file.closed:
+                temp_file.close()
+            os.remove(temp_file.name)
```
danieldk, line 199: Same here, these can raise, if that happens, we should continue removing temporary files.

curated_transformers/tests/repository/test_hf_hub.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+
+from curated_transformers.repository import HfHubRepository
+
+TEST_LINES = ["Line 1", "Line 2 🐍", "🤖" "\n\n", ""]
+
+
+@pytest.mark.slow
```
danieldk, line 8: I think we need another mark. We still want people to be able to run slow tests, even if they don't have access to our repo.

curated_transformers/tokenizers/tokenizer.py
```diff
@@ -337,8 +338,16 @@ def from_hf_hub_to_cache(
...
     def from_repo(cls: Type[Self], repo: Repository) -> Self:
+        def validate_tokenizer_file(tokenizer_file: RepositoryFile):
+            # Since repository files are loaded lazily, we need to
+            # open them to check their validity. Will throw if invalid.
+            with tokenizer_file.open(mode="rb") as f:
+                pass
+
         repo = TokenizerRepository(repo)
         tokenizer_file = repo.tokenizer_json()
+        validate_tokenizer_file(tokenizer_file)
```
danieldk, line 349: Hmmm, this does make an additional request/has some overhead.

danieldk (commented):
🎉 

Added some comments

curated_transformers/repository/hf_hub.py
```diff
@@ -31,26 +112,201 @@ def __init__(self, name: str, *, revision: str = "main"):
...
+    def repo(self) -> Repository:
+        return self._repo
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        if exc_type is None:
+            self._upload_temp_files()
+        self._release_temp_files()
```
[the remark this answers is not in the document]
shadeMe: Done.

curated_transformers/repository/hf_hub.py
shadeMe: Done.

curated_transformers/tests/repository/test_hf_hub.py
shadeMe: Added a new marker for upload tests.

curated_transformers/tokenizers/tokenizer.py
shadeMe: Rewrote the initializer to use a single request.

curated_transformers/repository/fsspec.py
```diff
@@ -97,15 +107,47 @@ def __init__(
...
+
     @property
     def _protocol(self) -> str:
         return self.fs.unstrip_protocol(self.repo_path)
+
+
+class FsspecTransactionContext(TransactionContext):
+    """
+    `fsspec`_ transaction context manager.
+    TODO: Implement/currently a noop
```
rmitsch, line 128: Should the TODO be in here? How is this a noop?

curated_transformers/repository/hf_hub.py
```diff
@@ -11,6 +14,84 @@
...
+        super().__init__()
+
+        self._repo = repo
+        self._remote_path = path
+        self._cached_path = lookup_file_in_cache(repo.name, repo.revision, path)
+
+    def _validate_model_repo(self):
+        api = HfApi()
+        # This will raise RepositoryNotFoundError/RevisionNotFoundError upon failure.
+        _ = api.model_info(repo_id=self._repo.name, revision=self._repo.revision)
```
rmitsch, line 44: Are you setting `_` on purpose instead of
```suggestion
        HfApi().model_info(repo_id=self._repo.name, revision=self._repo.revision)
```
?

curated_transformers/repository/hf_hub.py
```diff
@@ -31,26 +112,207 @@ def __init__(self, name: str, *, revision: str = "main"):
...
+            self._release_temp_files()
+
+    def _upload_temp_files(self):
+        if len(self._file_mappings) == 0:
+            return
+
+        # Flush buffers.
+        for temp_file in self._file_mappings.values():
+            if not temp_file.closed:
+                temp_file.flush()
```
rmitsch, line 178: Why is this necessary?

curated_transformers/repository/hf_hub.py
```diff
@@ -31,26 +112,207 @@ def __init__(self, name: str, *, revision: str = "main"):
...
+        # Collate all files as a single commit.
+        commit_ops = [
+            CommitOperationAdd(path_in_repo=remote_path, path_or_fileobj=temp_file.name)
+            for remote_path, temp_file in self._file_mappings.items()
+        ]
+        filenames = "\n".join(self._file_mappings.keys())
+        commit_message = f"Uploading files:\n{filenames}"
+
+        api = HfApi()
+        commit_info = api.create_commit(
```
rmitsch, line 189:
```suggestion
        HfApi().api.create_commit(
```
`commit_info` isn't used, right?

rmitsch (commented):
I'm not familar with the codebase yet, so my comments are somewhat superficial. Overall this looks reasonable to me. I'll do a second pass after the first batch of comments has been addressed.

curated_transformers/repository/file.py
```diff
@@ -40,6 +43,14 @@ def path(self) -> Optional[str]:
         """
         ...
 
+    @abstractmethod
+    def exists(self) -> bool:
+        """
+        Returns if the file exists. This can cause the file
+        to be cached locally.
+        """
```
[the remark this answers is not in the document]
shadeMe: We use [sphinx](https://www.sphinx-doc.org/en/master/) for our docs in this codebase, so the return type is automatically generated. We also don't explicitly mention the return value for property-like methods.

curated_transformers/repository/fsspec.py
shadeMe: Yeah, it's intended. The transaction semantics of `fsspec` are different from how we handle it for HF repos, so we need to think about it a bit more. The stub class is just there to aid the test suite/CI.

curated_transformers/repository/hf_hub.py
shadeMe: Yeah, it's to indicate that we are intentionally discarding the return value.

curated_transformers/repository/hf_hub.py
shadeMe: More of a sanity check as the user can consume the `IO`-like object outside of a `with` context.

curated_transformers/repository/hf_hub.py
shadeMe: True, but I'll `_` it to indicate that we're discarding the result (we might need it later).

PR #358, opened 2023-12-30
Added ELECTRA as a thin wrapper around BERT

KennethEnevoldsen:
## Description

Added support for loading ELECTRA model from the HF hub using a thin wrapper around BERTEncoder (required as a few keys need to be mapped differently) and re-using the same config.

The tests are a bit extensive (loading three different models); I imagine you might want to create your own dummy Electra models for testing.

## Checklist


curated_transformers/models/electra/_hf.py
```diff
@@ -0,0 +1,103 @@
...
+        "embeddings_project.bias", "embeddings.projection.bias"
+    ),
+    StringTransformations.replace(
+        "embeddings_project.weight", "embeddings.projection.weight"
+    ),
+]
+
+HF_CONFIG_KEYS: List[Tuple[HFConfigKey, Optional[HFConfigKeyDefault]]] = [
+    (CommonHFKeys.ATTENTION_PROBS_DROPOUT_PROB, None),
+    (CommonHFKeys.EMBEDDING_SIZE, None),
```
KennethEnevoldsen, line 75: I had to add embedding size here, which led to a conflict with the BERT model. This was what led to the thin wrapper class. It is def. possible to avoid it by implementing an if else logic in `_config_from_hf`.

This seemed like a reasonable comprise between not duplicating functionality and avoid coupling, though I could imagine you would want ELECTRA to be a part of BERT or completely independent.

curated_transformers/tests/models/electra/test_encoder.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.mark.skipif(not has_hf_transformers, reason="requires huggingface transformers")
+@pytest.mark.parametrize("torch_device", TORCH_DEVICES)
+@pytest.mark.parametrize("with_torch_sdp", [False, True])
+@pytest.mark.parametrize(
+    "model_name",
+    [
+        "jonfd/electra-small-nordic",
+        "Maltehb/aelaectra-danish-electra-small-cased",
+        "google/electra-small-discriminator",
+    ],
```
KennethEnevoldsen, line 21: I checked using a variety of models, but I imagine you might want to replace this with a dummy Electra model.

curated_transformers/tests/tokenizers/legacy/test_electra_tokenizer.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.mark.skipif(not has_hf_transformers, reason="requires huggingface transformers")
+@pytest.mark.parametrize(
+    "model_name", ["jonfd/electra-small-nordic", "Maltehb/aelaectra-danish-electra-small-cased", "google/electra-small-discriminator"]
+)
+def test_from_hf_hub_equals_hf_tokenizer(model_name: str, sample_texts):
+    compare_tokenizer_outputs_with_hf_tokenizer(
+        sample_texts, model_name, BERTTokenizer
+    )
```
KennethEnevoldsen, line 18: This test is simply to show/test that the electra models can use the BERT tokenizer

danieldk:
Awesome, thanks a lot! I hope to have some time over the coming days to review your PR.

curated_transformers/tests/models/electra/test_encoder.py
```diff
@@ -0,0 +1,29 @@
...
+)
+def test_encoder(model_name: str, torch_device, with_torch_sdp):
+    assert_encoder_output_equals_hf(
+        ELECTRAEncoder,
+        "jonfd/electra-small-nordic",
```
danieldk, line 26:
```suggestion
        model_name,
```

danieldk (commented):
Thanks a lot! Added a small comment.

curated_transformers/tests/models/electra/test_encoder.py
KennethEnevoldsen: Ahh yes of course!

KennethEnevoldsen:
Hi @danieldk I see that it now causes an error. Sadly I can't see the error on buildkite and when I run it locally I can't reproduce the error.

btw. I have also added a related PR over on spacy-curated-transformers (which I plan to finish up once this one is through)

danieldk:
@KennethEnevoldsen would you still like to work on this PR? Otherwise, I can also do the last bits to push it over the finish line.

KennethEnevoldsen:
Hi @danieldk I would love to. I have a deadline this week, but I can come back to it next week (potentially before if I find the time).

KennethEnevoldsen:
@danieldk I have fixed the isort issues and formatted using black. I have run the tests locally as well and they pass (or are skipped).

.github/workflows/test.yml
```diff
@@ -23,8 +23,8 @@ jobs:
           python -m black curated_transformers examples --check
       - name: isort
         run: |
-          python -m pip install isort -c requirements.txt
```
danieldk, line 26: I think something went wrong here.

KennethEnevoldsen:
@danieldk checked the error, but it does not seem to be related to the PR (in tokenizer test on rotary embedding). Seems like there is an additional argument seq_len.

danieldk:
> @danieldk checked the error, but it does not seem to be related to the PR (in tokenizer test on rotary embedding). Seems like there is an additional argument seq_len.

I just fixed the errors in #367, merging in `main` into this PR should fix the test issues.

danieldk:
Thanks a lot, looks good! I'll make a small test model in a bit to replace the tests.

PR #360, opened 2024-02-08
Clear output of Torch SDPA for masked pieces

danieldk:
## Description

Since Torch 2.1, the Torch memory-efficient SDPA GPU kernel [returns NaN](https://github.com/pytorch/pytorch/issues/110213) for pieces that are completely masked out. This leads to NaN propagation in the next attention layer, because masked pieces get an attention of zero, but zero times NaN is still NaN.

In this we fix this by setting masked tokens to zero to clear out any NaNs.

We currently rely on the query dimension of the mask to be singular, but in the future we should probably redesign the `AttentionMask` class to account for the differences between attention masks and causal masks.

## Checklist

### DIFF
--- a/curated_transformers/layers/attention.py
+++ b/curated_transformers/layers/attention.py
@@ -764,7 +764,9 @@ class ScaledDotProductAttention(AttentionScorer):
             #
             # Doing this properly requires a redesign of our AttentionMask
             # class.
-            assert attention_mask.bool_mask.size(-2) == 1, "Torch SDP does not support attention masks with non-broadcastable query length yet"
+            assert (
+                attention_mask.bool_mask.size(-2) == 1
+            ), "Torch SDP does not support attention masks with non-broadcastable query length yet"
             return torch.where(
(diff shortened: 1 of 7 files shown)


curated_transformers/layers/attention.py
```diff
@@ -712,27 +715,59 @@ def forward(
...
+            #
+            # So we can only do this when we have attention masks where
+            # the query length it not specified, which are typically 'pure'
+            # attention masks (not causal maskes or combined masks):
+            #
+            # [batch_len, 1, 1, key_len]
+            #
+            # Doing this properly requires a redesign of our AttentionMask
+            # class.
+            assert attention_mask.bool_mask.size(-2) == 1
```
shadeMe, line 767: We should add a message here as it's not an invariant at the moment.

```suggestion
            assert attention_mask.bool_mask.size(-2) == 1, "Torch SDP does not support attention masks with non-broadcastable query length yet"
```

PR #375, opened 2024-04-17
Fix activation lookup with Python 3.12.3

danieldk:
## Description

We used the metaclass `EnumMeta`/`EnumType` to override reporting of missing enum values (to give the full set of supported activations). However, in Python 3.12.3, the default value of the `name` parameter of `EnumType.__call__` method was changed from `None` to `_not_given`:

https://github.com/python/cpython/commit/d771729679d39904768f60b3352e02f5f491966c

Even though this is a public API (which now uses a private default value), it seems too risky to continue using it. So in this change, we implement `Enum.__mising__` instead for the improved error reporting.

## Checklist

### DIFF
--- a/curated_transformers/layers/activations.py
+++ b/curated_transformers/layers/activations.py
@@ -1,5 +1,5 @@
 import math
-from enum import Enum, EnumMeta
+from enum import Enum
 from typing import Type
 
 import torch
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
LGTM! Is there a unit test for this behaviour?

curated_transformers/layers/activations.py
```diff
@@ -71,6 +32,14 @@ class Activation(Enum, metaclass=_ActivationMeta):
     #: Sigmoid Linear Unit (`Hendrycks et al., 2016`_).
     SiLU = "silu"
 
+    @classmethod
+    def _missing_(cls, value):
```
svlandeg, line 36: I actually think this is more elegant than how we had it before :-)

danieldk: Yeah, somehow I missed (maybe pun intended?) this before.

danieldk:
> LGTM! Is there a unit test for this behaviour?

Not yet, but it sounds good to have this. I'll do a PR for that after getting new releases for 2.0.x and 1.3.x out.

